PR #11, opened 2017-07-01
Fix formatting bug and show stats when use make

faustinoaq:
Hi @kofno !

I found some issues while using Scry:

- Formatter don't replace all when old document is bigger that new one (Fixed with this PR)
- Sometimes Diagnostics are keep even when code is correct (Strange bug, I didn't know how to fix it)

Also I added `-s` flag to show stats when using make.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 darwin:
 	/bin/bash -c "mkdir -p bin/darwin"
-	/bin/bash -c "crystal build src/scry.cr --release --no-debug -o bin/darwin/scry"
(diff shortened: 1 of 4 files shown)


kofno:
Looks like this PR broke the build. Can you fix?

Also, I'm going out of town this week, but if there are still issues, I can take look when I get back. Maybe create a failing test case in another PR?

faustinoaq:
@kofno Don't worry, I'm going to my final university exams this week, I'll not have time to see code until the next 2 weeks. :sweat_smile: 

See you :smile:

PR #12, opened 2017-07-10
Improves performance when analyzing on save.

faustinoaq:
Hi @kofno I did some improves.

- ~~Verify that document text is different before analyze.~~
- Ensure to execute Garbage Collector. (free some MB)
- Avoid to check files shards and crystal lib.

### DIFF
--- a/src/scry/analyzer.cr
+++ b/src/scry/analyzer.cr
@@ -5,22 +5,23 @@ require "./publish_diagnostic"
 
 module Scry
   struct Analyzer
-    @@source = [] of String
-
     def initialize(@workspace : Workspace, @text_document : TextDocument)
       @diagnostic = PublishDiagnostic.new(@workspace, @text_document.uri)
     end
 
     def run
-      if @@source != @text_document.text
-        @@source = @text_document.text
-        @@source.map do |text|
-          unless @text_document.filename.starts_with?("/usr/lib") ||
-                 @text_document.filename.starts_with?("#{@workspace.root_uri}/lib")
-            analyze(text)
-          end
-        end.flatten.uniq
+      @text_document.text.map do |text|
+        unless inside_path?
+          analyze(text)
+        end
+      end.flatten.uniq
+    end
+
+    private def inside_path?
+      ENV["CRYSTAL_PATH"].split(':').each do |path|
+        return true if @text_document.filename.starts_with?(path)
       end
+      false
     end
 
     # NOTE: compiler is a bit heavy in some projects.
--- a/src/scry/analyzer.cr
+++ b/src/scry/analyzer.cr
@@ -5,12 +5,22 @@ require "./publish_diagnostic"
 
 module Scry
   struct Analyzer
+    @@source = [] of String
+
     def initialize(@workspace : Workspace, @text_document : TextDocument)
       @diagnostic = PublishDiagnostic.new(@workspace, @text_document.uri)
     end
 
     def run
-      @text_document.text.map { |t| analyze(t) }.flatten.uniq
+      if @@source != @text_document.text
+        @@source = @text_document.text
+        @@source.map do |text|
+          unless @text_document.filename.starts_with?("/usr/lib") ||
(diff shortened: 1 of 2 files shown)


src/scry/analyzer.cr
```diff
@@ -5,12 +5,22 @@ require "./publish_diagnostic"
 
 module Scry
   struct Analyzer
+    @@source = [] of String
```
kofno: I've been out of the Scry codebase for a while, but I'm a bit concerned about this being set as a class var. Doesn't this make it impossible to check a project with multiple files against a single running server instance?

src/scry/analyzer.cr
```diff
@@ -5,12 +5,22 @@ require "./publish_diagnostic"
...
     def initialize(@workspace : Workspace, @text_document : TextDocument)
       @diagnostic = PublishDiagnostic.new(@workspace, @text_document.uri)
     end
 
     def run
-      @text_document.text.map { |t| analyze(t) }.flatten.uniq
+      if @@source != @text_document.text
+        @@source = @text_document.text
+        @@source.map do |text|
+          unless @text_document.filename.starts_with?("/usr/lib") ||
```
kofno: Is this a global shard home directory? Is there an env var we could use instead?

src/scry/implementations.cr
```diff
@@ -39,6 +39,10 @@ module Scry
         end
         ResponseMessage.new(@text_document.id, locations)
       end
+    rescue ex
+      nil
```
kofno: Should we at least log that an exception happened? What kind of exceptions are we expecting here?

src/scry/analyzer.cr
faustinoaq: In VSCode, every workspace launch a new instance of scry, in NVIM too. I don't know in Atom, however if the code inside `text_document.text` changes, then the source file is updated.

My goals with this variable is to check that source is different when document is saved, because the actual behavior is checking the file too much, spending memory and CPU time.

Is a class variable because a new instance of Analyzer is created  in a new message, and I didn't have other way to handle this.

src/scry/analyzer.cr
faustinoaq: Yes I can use, `ENV["CRYSTAL_PATH"]`
Let me fix that, I will do on weekend.

src/scry/implementations.cr
faustinoaq: Yes something like, `Log.logger.error("A error was found while searching definitions\n#{ex}")`

faustinoaq:
@kofno Thanks you for review my code :sweat_smile: 

I will improve this PR on weekend.

The aims of this PR is to reduce scry resources, because currently scry analyzer take up to 500 Mb with a simple file.

Without analyzer scry uses only 10Mb in my computer to do formatting, syntax checking and goto defitiniton.

I know that isn't a scry problem but a crystal problem, but at least we can do something for now.

Cheers

src/scry/analyzer.cr
kofno: We can leave this for now, but this seems like a likely place for a race condition. It may be time to consider re-architecting the Analyzer.

faustinoaq:
Yes, I think the same, Maybe the analyzer could be implemented using spawn
and channels.
No problem with leaving it out, I'll change this in the next weeks.

kofno:
I want to read of the code for some of the the other language servers, just to get some ideas before we get too deep into refactoring. I'd also like to read some general Crystal and Go server code to get a good feel for using channels again. I worked through some examples before, but I need a refresher I think.

faustinoaq:
Hi @kofno :smile: 

I removed `@@souce` class variable, and now it uses `CRYSTAL_PATH` to avoid check files inside crystal lib.

faustinoaq:
@kofno can you check this, please. I want to delete my fork because I made some [mistakes ](https://github.com/faustinoaq/scry/commit/e6729270f3e1a137772b18d1e1b44f8e2705154a) :sweat_smile: while I was merging in other branches, so I want to create a new clean fork from this repo.

kofno:
Done! Thanks again!

PR #15, opened 2017-07-22
Configure Travis to publish Linux and macOS builds

### LINKED ISSUE #3: [RFC] Better distribution story for scry
VSCode packages can't be larger then 20MB. So I can't distribute Cystal-IDE w/ bundled binaries for all platform/arch combos we'll want to support (probably a bad idea anyway).

So how should we distribute scry?

1) Maintain brew taps and apt-get repos. Brew taps don't seem that hard to maintain, but I'm uncertain of the cost in effort for keeping an apt distribution. And that excludes arch and redhat linux users.

2) Build logic into Crystal-IDE to fetch proper version of scry (and detect updates). There's a lot of hidden effort here; like prompting users that an update is available, etc.

3) See if it's possible to make scry distributable as a development dependency. This is how, for example, rake is distributed most often to ruby projects. This would actually make the most sense, since you may need a different scry on different projects. However, I'm not certain that shards supports distributing binary dependencies in this way.

keplersj:
This does abandon the Docker solution you were using before. However using the virtualized environment Travis CI provides is more inline with the default implementation being worked on for Crystal projects on Travis. Sadly, at the moment the default Crystal config for Travis does not work with this project. This is being worked on though (ref: travis-ci/travis-build#1117).

This setup installs Crystal on Ubuntu and macOS, runs the specs, and when successful builds a tar of the compiled scry binary. If running on a tagged commit Travis publishes that tarred binary to the tagged release on the GitHub Releases page. This encourages version tagging and maintaining a living Changelog on the GitHub Releases page.

I did have to disable a spec as it was dependent on a Travis environment constant that was no longer correct.

I'm happy to answer any questions or make any changes.

Fixes #3

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,62 @@
 language: generic
+
+matrix:
+  include:
+    - os: linux
+      dist: trusty
+      deploy:
+        file:
+          - scry_linux.tar.gz
+    - os: osx
+      deploy:
+        file:
+          - scry_macOS.tar.gz
+
 sudo: required
-services:
-- docker
-before_install:
-- docker build -t kofno/scry .
+
+install:
+- |
+  set -e
+  set -x
+  case $TRAVIS_OS_NAME in
+  linux)
+    curl https://dist.crystal-lang.org/apt/setup.sh | sudo bash
+    sudo apt-get install -y crystal libgmp-dev llvm-3.8-dev build-essential
+    ;;
+  osx)
+    rm -f /usr/local/bin/shards
+    brew update
+    brew install crystal-lang
+    brew link llvm --force
+    ;;
+  *)
+    exit 1;;
+  esac
+
+  set +x
+
 script:
-- docker run kofno/scry /bin/sh -c "cd /opt/scry; crystal spec"
+- crystal spec --error-trace --stats --progress
+
+after_success:
+- |
+  set +e
+  set -x
+  case $TRAVIS_OS_NAME in
+  linux)
+    travis_wait make linux
+    tar -cvzf scry_linux.tar.gz bin/linux/scry
+    ;;
+  osx)
+    travis_wait make darwin
+    tar -cvzf scry_macOS.tar.gz bin/darwin/scry
+    ;;
+  *)
+    exit 1;;
+  esac
+
+  set +x
+
 notifications:
   webhooks:
     urls:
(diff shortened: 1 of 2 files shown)


keplersj:
Resolves #3

kofno:
This looks good to me. Maybe I'll leave it open a bit to see if @faustinoaq has any questions.

faustinoaq:
Hi, I'm okay with this PR, doing automatic builds is awesome, but `spec` is necessary. I propose that every change must do a new PR and show a screenshot passing specs since Travis wouldn't do that. Showing screenshots Like this:

PR with spec screenshot

Maybe you can use a PR template for contributions.

@kofno and @keplersj something I'm thinking about Scry distribution is that actual configuration is bind to `crystal-ide`, but Scry could be used by Atom, NVIM and SublimeText not only VSCode.

I removed `scry.7z` from [vscode-crystal-lang](https://github.com/faustinoaq/vscode-crystal-lang/blob/master/CHANGELOG.md#024---2017-07-20) to use these builds, and this scry is working well, I just get an error in `initialize` method, because my extension doesn't have `crystal-ide` on their settings.

I suggest to remove/reduce client configuration for the Scry server:

```crystal
require "json"

module Scry
  # Language Server Protocol doesn't specify settings type
  # because it depend of client.
  # See: https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#didchangeconfiguration-notification
  #
  # Replace the key to match your client config.
  struct Settings
    JSON.mapping({
      crystal_config: {type: Customizations, key: "crystal-ide"},  # => Change this to be generic
    })
  end

  # Currently configured for vscode-crystal-ide
  # ---------------------------------------------------------------
  # Proposal: Reduce Client/Server configuration to be more generic
  # ---------------------------------------------------------------
  struct Customizations
    JSON.mapping({
      max_number_of_problems: {type: Int32, key: "maxNumberOfProblems"},
      backend:                String,
      custom_command:         {type: String, key: "customCommand"},
      custom_command_args:    {type: Array(String), key: "customCommandArgs"},
      log_level:              {type: String, key: "logLevel"},
    })
  end
end
```

Changing customization to be more generic would help to editors to have the best support by Scry.

Avoiding something like this:

screenshot_20170724_090902

Perhaps, not removing but using scry arguments or something similar.

.travis.yml
```diff
@@ -1,24 +1,75 @@
...
+    ;;
+  *)
+    exit 1;;
+  esac
+
+  set +x
+
 script:
-- docker run kofno/scry /bin/sh -c "cd /opt/scry; crystal spec"
+- crystal spec --debug --error-trace --stats --progress
```
faustinoaq: Why `--debug` flag?

keplersj:
@faustinoaq I'm confused.

> doing automatic builds is awesome, but spec is necessary.

Travis still runs the specs, it just doesn't run that one spec which would've failed on local environments anyway. In this new setup Travis runs the specs, and when successful builds the binary. From there if the commit being tested is a tagged commit Travis uploads the artifact to GitHub Releases.

> I suggest to remove/reduce client configuration for the Scry server

That all sounds good. I never had any issue with it but perhaps the Atom client doesn't enforced those settings at the moment. Also, a new issue should be opened to discuss this topic - it's kinda cluttering this one.

.travis.yml
keplersj: Honestly, I was just throwing on parameters that I thought might be helpful in case of failure. You're right though that one doesn't make much sense - removing it!

faustinoaq:
> @faustinoaq <https://github.com/faustinoaq> I'm confused.

My bad, I misunderstood the issue :sweat_smile:

faustinoaq:
Travis is passing on Linux, but in MacOS show the following:

```
==> Installing crystal-lang dependency: llvm
==> Downloading https://homebrew.bintray.com/bottles/llvm-4.0.1.el_capitan.bottl
Error: Failed to download resource "llvm"
```

Is llvm-4.0.1 available on Homebrew?

keplersj:
It is. The reason Travis failed on macOS in that commit is to do a timeout.

> No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

If @kofno reruns the build it should be fine.

faustinoaq:
`"check implementation response content on Travis CI"` This spec does not have reproducibility,  not very useful. I did it because I wanted to test implementations on Travis, maybe IO::Memory is a better way.

For know I think we can left away this implementation test and activate automatic releases.

What do you think guys?

keplersj:
Hey @kofno if everything looks good can we get this merged and push a compiled version out?

PR #19, opened 2017-08-09
Update link to Atom package.

keplersj:
This package will make use of scry in its next update.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ The server is implemented in Crystal.
 
 Scry will be distrubuted as part of editors extensions, see:
 
-- [atom-scry](https://github.com/atom-crystal/scry)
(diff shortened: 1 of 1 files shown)


faustinoaq:
Hi @keplersj I think would be useful update the references in source code too.

:point_right: https://github.com/kofno/scry/blob/master/src/scry/protocol/settings.cr#L41

keplersj:
@faustinoaq I agree it should be kept up to date, but the Atom package doesn't make use of those right now anyway. Would it be better to keep the commented code up to date or simply remove the commented code.

faustinoaq:
@keplersj the [LSP specifies settings](https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#didchangeconfiguration-notification) as `any`. So, if atom extension doesn't need settings I think is safe to remove it.

PR #31, opened 2017-10-05
Remove id from NotificationMessage

bmulvihill:
According to LSP 3.0 spec regarding [NotificationMessages](https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#notification-message) the id field is not part of the spec.  This PR removes the optional id from the NotificationMessage type

### DIFF
--- a/src/scry/protocol/notification_message.cr
+++ b/src/scry/protocol/notification_message.cr
@@ -18,7 +18,6 @@ module Scry
   struct NotificationMessage
     JSON.mapping({
       jsonrpc: String,
-      id:      Int32?,
       method:  String,
       params:  NotificationType,
     }, true)


faustinoaq:
Hi @bmulvihill I think we can merge this when other editors support LSP 3.0 

Thanks you for you contribution!

bmulvihill:
@faustinoaq Does it break < 3.0?, I don't see the id field in LSP 2.0 or 1.0 specs

faustinoaq:
Oh, @bmulvihill you're right! I checked LSP 1.0 and 2.0 and they didn't have id field on `NotificationMessage` spec

So @keplersj Can we merge it ?

faustinoaq:
Thanks you! @bmulvihill

PR #33, opened 2017-10-22
Adds initialized notification

bmulvihill:
Allows Scry to receive the [initialized notification](https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#initialized). 

Adds new param type `VoidParams` which is an empty JSON object used by the initialized notification


faustinoaq:
@bmulvihill Thanks you! 👍

Issue #35, opened 2017-10-22
Scry formatting bug


faustinoaq:
On ST3 to reproduce this issue  try scry formatting disabling crystal `auto_format` on crystal package setting:

```json
{
  "auto_format": false,
  "crystal_cmd": "crystal"
}
```

![bug01](https://user-images.githubusercontent.com/3067335/31861119-da2fafd2-b6ec-11e7-8db3-8544bdedc4d2.gif)

As you can see scry formatting is almost deleting all file text.

Ref: #34

bmulvihill:
@faustinoaq Did some digging on this - I believe it is due to the way Sublime LSP sets its text points.  So we send the our response which contains different ranges.  These ranges get converted into Sublime Regions. Sublime is overlapping the regions it creates because all of our text edits have the same MAX value, which essentially converts them all into one region (This is why we are only seeing the first line get formatted )https://github.com/crystal-lang-tools/scry/blob/master/src/scry/formatter.cr#L34

I looked at how Go's Language server does it and it looks like they send back a single text edit for the entire format https://github.com/sourcegraph/go-langserver/blob/master/langserver/format.go#L53.  I did the same on our and it seems to solve the problem, since we only send back a single Range and Sublime uses a single region.  Is there any particular reason we would need to send back each individual TextEdit since we end up formatting the entire document anyway?

PR #10, opened 2017-06-30
Scry support for formater,  GoTo definition and other updates

faustinoaq:
Hi @kofno!

I made some changes to improve your project, allowing to use with other plugins and editors. By example, this PR includes [ResponseError](https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#response-message) and Position classes made by @keplers for [atom-scry](https://github.com/atom-crystal/scry).

By default is configured for `crystal-ide` but It's easy to configure for others extensions like [vscode-crystal-lang](https://github.com/faustinoaq/vscode-crystal-lang#server-new).

I added more features to Scry. The features are:

- Formatter
- GoTo Definition

There is a screen capture:

ezgif-2-0723b7d6d7

Also, I refactored some files to improve the structure of [Language Server Protocol](https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md) specification.

Some screenshots:

screenshot_20170629_221711

screenshot_20170629_221618

screenshot_20170629_210335

screenshot_20170628_115314

Finally I added a logo to your project to be more developer friendly :smile: 

tictft8 - imgur

> Note: Scry is compatible with "vscode-languageclient": "2.2.1" most recent client will need some changes in this server.

PR #37, opened 2017-10-24
Send back single formatted TextEdit

bmulvihill:
Fixes https://github.com/crystal-lang-tools/scry/issues/35

- Formats document with a single TextEdit rather than many since we end up formatting the entire source anyway

### DIFF
--- a/spec/spec_helper.cr
+++ b/spec/spec_helper.cr
@@ -26,7 +26,7 @@ module Scry
 
   BUILD_ERROR_EXAMPLE = %({"file":"/home/aa.cr","line":4,"column":1,"size":1,"message":"Oh no!, an Error"})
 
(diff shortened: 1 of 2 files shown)


faustinoaq:
@bmulvihill Thanks you!

This bug is my fault 😅 

I tried to fix a bug, related to files above 10000 characters, I thought this was because I was using a single TextEdit, so I implemented multiples Textedit <sup>1</sup> Then I found I was using `content.size` instead of `content.bytesize`  <sup>2</sup> but my bad implementation kept in codebase 😅 

1. https://github.com/crystal-lang-tools/scry/blob/8a37e842f77b543337f6a19019419a2203cac719/src/scry/formatter.cr#L13

2. https://github.com/crystal-lang-tools/scry/pull/10/commits/57dad798cc481951bfab89d5f18652bc02ff4966#diff-786bcb9064e3287ebb531cf94135c945R28

Issue #38, opened 2017-10-25
Support list symbols


faustinoaq:
I think Scry can support  list [Document](https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#textDocument_documentSymbol) and [Workspace](https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#workspace_symbol) symbols using [Crystal Ctags](https://github.com/SuperPaintman/crystal-ctags) by @SuperPaintman

![crystal-ctags](https://raw.githubusercontent.com/SuperPaintman/crystal-ctags/master/README/screenshot.png)

Usecase: https://github.com/crystal-lang-tools/vscode-crystal-lang/issues/24

faustinoaq:
Fixed by #39

PR #43, opened 2018-01-14
Add property and variable symbols

bmulvihill:
Adds property and variable symbols to Scry, in line with how `vscode-crystal-lang` defines them.

Related to #38

### DIFF
--- a/spec/scry/symbol_spec.cr
+++ b/spec/scry/symbol_spec.cr
@@ -9,14 +9,14 @@ module Scry
     end
 
     it "contains SymbolInformation" do
-      text_document = TextDocument.new("uri", [""])      
+      text_document = TextDocument.new("uri", [""])
       processor = SymbolProcessor.new(text_document)
       response = processor.run
       response.result.is_a?(Array(SymbolInformation)).should be_true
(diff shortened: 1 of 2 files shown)


keplersj:
The code looks good. I just want to see the CI pass before I okay this. Just restarted the CI jobs - will see if they are successful in a few minutes.

keplersj:
Hmm... Both errors on first glance look like unrelated LLVM errors. Did Crystal update its LLVM dependency while I wasn't looking and now our CI config needs to be updated?

faustinoaq:
> while I wasn't looking and now our CI config needs to be updated?

@keplersj Maybe we should remove this line 👉 https://github.com/crystal-lang-tools/scry/blob/master/.travis.yml#L3 and use LLVM 4.0 or even 5.0, WDYT?

keplersj:
As long as it works it's good with me

faustinoaq:
@bmulvihill I tested it an works very well :tada: 

vokoscreen-2018-02-25_01-01-23

I gonna merge this, Thank you so much for collaborate to scry! :+1:

Issue #28, opened 2017-09-28
Support auto completion


faustinoaq:
Currently @TechMagister have a really good tool for autocompletion [0]

I think we can implement auto completion on Scry using cracker [0]

[0] - https://github.com/TechMagister/cracker

### Update

Adding a TODO list by @laginha87 

- [x] Add dependency graph, simple method db and method completion with the basic scenarios.
- [x] Add method resolve for documentation
- [ ] Add the class graph generation and add super method calls. Should handle namespaced classes better.
- [ ] Improve autocompletion by adding more scenarios
- [ ] Add support for module inclusion

laginha87:
Hi, I wanted to contribute and this seem like a good place to start. 
The description is a bit vague on the implementation and wanted to check whats the best path to take. 
1. Add cracker as a shard dependecy, and use the classes. 
2.  Add the cracker bin and just use it. 
3.  Copy over the relevant source code.

faustinoaq:
> 3. Copy over the relevant source code.

I think cracker hasn't been updated almost a year ago, so its code could be incompatible right now. I propose to copy the relevant code and create a scry module based on cracker source.

@kofno @keplersj @bmulvihill WDYT?

keplersj:
I say we should view `cracker` as prior art in the field of providing autocompletion for Crystal. We can build autocompletion functionality into `scry` without it having anything to do with `cracker`. We don't need to build `cracker` functionality into `scry` - we need autocompletion functionality built into `scry`.

TechMagister:
Cracker have its own server, so using it with scry involve having 2 servers running, which is probably not what you want.

keplersj:
@TechMagister Would you like to join the team and help build autocompletion into `scry`?

TechMagister:
Sadly, I have to decline your invitation, I have not enough time to develop something significant. But I'll follow your work with attention.

laginha87:
I've been coding  a lot on getting  method autocompletion to work, I have a branch right now with a bunch of stuff working already but it seems like its doing to much so I'm going to split up the work in smaller more focused branches.
Wanted to explain the approach I used which is based on cracker.

The basic idea of the method autocomplete is to:
1) Preload all methods by class on startup on a method DB
2) When Autocomplete is triggered try to identify the type of the object that owns the method through regexes
3) Get the methods for the type that was found from the DB

The approach I'm going with is to:
[...]

PR #44, opened 2018-02-01
Try to fix Travis builds

faustinoaq:
This PR is for updating and trying to fix Travis builds

PR #46, opened 2018-02-07
Add autocompletion to module imports.

laginha87:
I set up the groundwork to add autocompletion to scry.
Autocompletion needs the files in memory and not the saved files so I had to add a bunch of code to handle that.
Since I'm not super familiar with writing visitors and stuff I though I'd add autocompletion to the require "" statements because scry's autcompletion can work at the same time as the vscode crystal extension. 

Related to #28


shard.yml
```diff
@@ -7,3 +7,7 @@ authors:
   - Faustino Aguilar <http://faustinoaq.github.io>
 
 license: MIT
+
+targets:
```
laginha87: I added this so scry can be built by doing ```shards build```

faustinoaq:
This is awesome! @laginha87 Thank you so much!!! 😄

faustinoaq:
I gonna take some time to review this, I'll also push a fix for Travis CI (started on https://github.com/crystal-lang-tools/scry/pull/44)

faustinoaq:
Hi @laginha87 I tested your PR and `require` completion works super nice :+1: 

vokoscreen-2018-02-25_00-58-51

However, something in this PR is breaking other features like word and snippet completion. Maybe this happens because is not full autocompletion yet and currently just works for module imports.

screenshot_20180225_010817

This PR is being awesome! :clap: :tada: :smile:

faustinoaq:
Oh! this works pretty nice with local modules as well :smile: 

vokoscreen-2018-02-25_01-31-34

faustinoaq:
Maybe we can find a way to include symbols on auto-completion, like I did [here](https://github.com/crystal-lang-tools/vscode-crystal-lang/blob/master/src/crystalCompletion.ts#L97)

Also, if no autocompletion is found, we can try to suggest words and snippets:

screenshot_20180225_015132

WDYT?

laginha87:
The issue was happening [here](https://github.com/laginha87/scry/blob/13d1ddf86143c72a26daaa46e055127eb71fd9b9/src/scry/completion_provider.cr#L32) I was raising an exception when I couldn't identify the kind of autocompletion to apply. The exception was uncaught and then vscode would wait for the response which would never show up.
I changed it to respond with an empty array of completion items when that happens, this way it includes the snippets and symbols from the extension.
I haven't looked at all of the spec for a language server but should snippets be part of it or should they come from the extension that connects to the language server ?

laginha87:
Another thing that may cause the Loading... issue is the lack of concurrency in scry. Because scry processes requests sequentially there are tasks that are run when you open a file that you to have to wait for before getting a response for autocompletion.

faustinoaq:
Hi @laginha87 I tried your last changes and looks like `Loading...` issue is fixed :+1: 

ezgif-4-585c247483

However, I modified your code a bit and seems like `CancelParams` isn't needed.

ezgif-4-e88804acb3

> I haven't looked at all of the spec for a language server but should snippets be part of it or should they come from the extension that connects to the language server ?

The extension code that connect to LSP is pretty simple, just [a few line of codes](https://github.com/crystal-lang-tools/vscode-crystal-lang/blob/master/src/crystalMain.ts#L50-L59).
I think it can't manage that behavior.

faustinoaq:
> Another thing that may cause the Loading... issue is the lack of concurrency in scry. Because scry processes requests sequentially there are tasks that are run when you open a file that you to have to wait for before getting a response for autocompletion.

Yeah, I thought that, so, I tried using concurrent channels: (I'll open a new PR about this)

```crystal
    next_request = Channel(Nil).new
    context = Context.new

    loop do
      spawn do
        content = Request.new(STDIN).read
        request = Message.new(content).parse
        results = context.dispatch(request)
      rescue ex
        results = [ResponseMessage.new(ex)]
      ensure
        response = Response.new([results].flatten)
        response.write(STDOUT)
        next_request.send nil
      end
      select
      when next_request.receive
        Log.logger.info("Scry has processed a request!")
        next
      else
        sleep 1
      end
    end
```

And `Loading...` issue was still happening because it was raising a `Completion` error.

Fixed when it returns an empty array.

faustinoaq:
BTW, Please, sync the master branch on this PR, I can't see update button, I don't know why :sweat_smile:

laginha87:
I updated the branch with master and fixed the tests.
The cancel request not being handled I don't think it causes issues but if we don't handle it it'll raise an exception and show an error in the developer tools.

spec/scry/symbol_spec.cr
```diff
@@ -19,39 +19,39 @@ module Scry
       text_document = TextDocument.new("uri", ["class Test; end"])
```
bmulvihill: Any reason we are updating symbol specs in this PR?  Seems unrelated.

laginha87: Because of the way response message https://github.com/crystal-lang-tools/scry/pull/46/files#diff-82346a49d1a2a5c31d9a3445d7a9127dR9  I had to add a single CompletionItem as a response for the completion resolver.
The symbol specs expect an array in the response.

src/scry/completion/require_module_context.cr
```diff
@@ -0,0 +1,37 @@
+module Scry::Completion
+  class RequireModuleContext < Context
+    RELATIVE_IMPORT_REGEX = /^\./
+
+    def initialize(@import : String, @text_document : TextDocument)
+    end
+
+    def find
+      case @import
+      when RELATIVE_IMPORT_REGEX
```
bew: `when .starts_with? '.'` would be faster here I think (and less regex is almost always better :smile: )

laginha87:
Thanks for all the feedback.
I applied the change to not use a regex.
Is this ready to be merged?

faustinoaq:
> Is this ready to be merged?

Yeah, I think it is ready, Thank you so much for this PR :smile: :+1: 

@keplersj I'm working on concurrent rpc for scry, I think we should publish a new release after that, WDYT?

keplersj:
@faustinoaq 👍. In truth I'd be okay with configuring something like [semantic-release](https://github.com/semantic-release/semantic-release) for scry.

PR #53, opened 2018-03-04
Make scry faster and lightweight

### LINKED ISSUE #30: Refactoring: embedded "compiler/crystal" by crystal commandline
Currently Scry uses a huge amout of memory because the crystal compiler is embedded on it.
Scry analyze files every time a file is saved without checking if the file have been truly changed increasing even more the footprint (aprox. over 300 MB RAM on Linux)

Also maybe we can use external `crystal` processes instead of using `require "compiler/crystal/**"` ('cause, compiler is very heavy) I think using processes will free memory easier and improve Scry performance.

Also new features as _types on hover_ and _refactoring_ could be easier to implement.

faustinoaq:
Hi, this PR implements lightweight features:

- Lightweight analyzer and implementations
- Avoid heavy compiler dependencies
- Use external process to free up memory

Before:

screenshot_20180304_083813

With this PR changes:

screenshot_20180304_083640

Same features and behaviors, more RAM friendly and lightweight binary `19.2MB` vs `4.7MB`

Oh, also compilation time has been reduced drastically:

```
$ time crystal build src/scry.cr -s -p --no-debug
Parse:                             00:00:00.000316202 (   0.25MB)
Semantic (top level):              00:00:00.404959721 (  52.33MB)
Semantic (new):                    00:00:00.002962556 (  52.33MB)
Semantic (type declarations):      00:00:00.032710793 (  52.33MB)
Semantic (abstract def check):     00:00:00.002463025 (  52.33MB)
Semantic (ivars initializers):     00:00:00.001851865 (  52.33MB)
Semantic (cvars initializers):     00:00:00.240542928 (  68.33MB)
Semantic (main):                   00:00:01.663311243 ( 196.77MB)
Semantic (cleanup):                00:00:00.006314623 ( 196.77MB)
Semantic (recursive struct check): 00:00:00.002034422 ( 196.77MB)
Codegen (crystal):                 00:00:01.197870686 ( 228.77MB)
Codegen (bc+obj):                  00:00:05.765841086 ( 228.77MB)
Codegen (linking):                 00:00:00.761140190 ( 228.77MB)
                                          
Codegen (bc+obj):
 - no previous .o files were reused
crystal build src/scry.cr -s -p --no-debug  14.42s user 0.81s system 149% cpu 10.159 total
```

Previously scry compilation used a huge amout of RAM (above 1GB) and took several minutes to compile in my pc :sweat: 

Fixes #30

Related comment on https://github.com/crystal-lang-tools/scry/pull/48#issuecomment-370222717

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ matrix:
         - DEPLOY_DIR=bin/darwin
 
 install:
-  - bin/ci  prepare_build
+  - bin/ci prepare_build
 
 script:
   - bin/ci build
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,6 @@
 FROM crystallang/crystal:0.24.1
 
 RUN apt-get update
-RUN apt-get install -y llvm-4.0-dev
 RUN apt-get clean
 RUN rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ Scry will be distributed as part of editors extensions, see:
 
 ## Development && Roadmap
 
-> **Note:** crystal and LLVM are required in order to compile scry.
+> **Note:** crystal is required in order to compile scry.
 
 Ongoing, in [our](https://github.com/kofno/scry#contributors) free time.
 
--- a/bin/ci
+++ b/bin/ci
@@ -65,8 +65,7 @@ prepare_build() {
   on_linux docker build -t crystal/scry-image .
 
   on_osx brew update
-  on_osx brew install llvm crystal-lang
-  on_osx brew link --overwrite --force llvm
+  on_osx brew install crystal-lang
 }
 
 
--- a/spec/scry/initialize_spec.cr
+++ b/spec/scry/initialize_spec.cr
@@ -6,7 +6,7 @@ module Scry
       initer = Initializer.new(
         InitializeParams.from_json({
           processId:    1,
-          rootPath:     "/homa/main/Projects/Experiment",
+          rootPath:     "/home/main/Projects/Experiment",
           capabilities: {} of String => String,
           trace:        "off",
         }.to_json),
--- a/src/scry/analyzer.cr
+++ b/src/scry/analyzer.cr
@@ -39,7 +39,7 @@ module Scry
     private def crystal_build(filename, source)
       code = IO::Memory.new(source)
       String.build do |io|
-        Process.run("crystal", ["build", "--no-color", "--error-trace", "-f", "json", filename], output: io, error: io, input: code)
+        Process.run("crystal", ["build", "--no-codegen", "--no-color", "--error-trace", "-f", "json", "--stdin-filename", filename], output: io, error: io, input: code)
       end
     end
   end
--- a/src/scry/analyzer.cr
+++ b/src/scry/analyzer.cr
@@ -1,4 +1,3 @@
-require "compiler/crystal/**"
 require "./workspace"
 require "./text_document"
 require "./publish_diagnostic"
@@ -26,17 +25,22 @@ module Scry
 
     # NOTE: compiler is a bit heavy in some projects.
     private def analyze(source)
-      source = Crystal::Compiler::Source.new(@text_document.filename, source)
-      compiler = Crystal::Compiler.new
-      compiler.color = false
-      compiler.no_codegen = true
-      compiler.debug = Crystal::Debug::None
-      compiler.compile(source, source.filename + ".out")
-      [@diagnostic.clean]
-    rescue ex : Crystal::Exception
-      @diagnostic.from(ex)
-    ensure
-      GC.collect
+      response = crystal_build(@text_document.filename, source)
+      if response.empty?
+        [@diagnostic.clean]
+      else
+        @diagnostic.from(response)
+      end
+    rescue ex
+      Log.logger.error("A error was found while searching diagnostics\n#{ex}")
+      nil
+    end
+
+    private def crystal_build(filename, source)
+      code = IO::Memory.new(source)
+      String.build do |io|
+        Process.run("crystal", ["build", "--no-color", "--error-trace", "-f", "json", filename], output: io, error: io, input: code)
+      end
     end
   end
 end
--- a/src/scry/implementations.cr
+++ b/src/scry/implementations.cr
@@ -1,5 +1,3 @@
-require "compiler/crystal/**"
-
 require "./log"
 require "./protocol/location"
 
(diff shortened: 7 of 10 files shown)


bew:
I wonder how the completion system will work with this

faustinoaq:
> I wonder how the completion system will work with this

@bew Completion feature is untouched

laginha87:
Completion just uses the compiler/syntax from crystal to build the ast trees.
Glad to see the reduction in compilation time. 
I don't how I feel about passing it off to the crystal tools though, seems like it might create a bigger dependency on what version of crystal the user is using.

faustinoaq:
> Completion just uses the compiler/syntax from crystal to build the ast trees. Glad to see the reduction in compilation time.

:+1: 

> I don't how I feel about passing it off to the crystal tools though, seems like it might create a bigger dependency on what version of crystal the user is using.

@laginha87 Then, a developer using scry will see the errors and implementation behaviors according to the crystal version that he/she is using. Will not depend on the crystal version that scry was compiled, so finally, is another benefit for using an external crystal process :smile:

keplersj:
Like I said [here](https://github.com/crystal-lang-tools/scry/issues/30#issuecomment-332900978) I'm still not sure this is a good idea.

If this PR included README changes stating that this is dependent on the user having the Crystal compiler installed in their PATH and included a mechanism to alert the user if this they don't when the server starts I would be more onboard with a change like this.

faustinoaq:
> If this PR included README changes stating that this is dependent on the user having the Crystal compiler installed in their PATH and included a mechanism to alert the user if they don't when the server starts I would be more onboard with a change like this.

Hi @keplersj :smile: 

We're already dependent on crystal command installed and avaliable on PATH. Scry is already executing `crystal` here:

https://github.com/crystal-lang-tools/scry/blob/1bff42ea0624eb991f11a7a90307a49fe4dce6cf/src/scry/environment_config.cr#L25

So currently if a user doesn't have crystal installed, then scry won't work

This PR just take advantage of using crystal command to be memory friendly, to compile faster and to generate a smaller binary.

keplersj:
@faustinoaq Hmm... Okay. I still can't shake this feeling that relying on the Crystal compiler command line is going to be a bad idea in the long run and probably breaks with the norms of other LSP servers. But, you have good points. I'm okay with this for now.

I do want to hear @bcardiff's opinion on this though.

faustinoaq:
> I still can't shake this feeling that relying on the Crystal compiler command line is going to be a bad idea in the long run and probably breaks with the norms of other LSP servers.

The `src/protocol` folder will work as usual, this PR just modifies the way to process 2 heavy features (diagnostics and implementations) and keeps other things untouched.

In fact the only disadvantage of using this approach is that compiler will outputs extra characters to stdio **in some special situations**: like `Using bin/crystal at commit kjdf6zx` (can be disabled) commonly shown in local crystal builds from git repo or benchmark warning like `Using benchmark without --release flag, bla, bla...` or some crystal bug. because it will break json output parsing.

bmulvihill:
@faustinoaq How will this affect CI?  We should be able to remove any LLVM dependencies?

faustinoaq:
> How will this affect CI? We should be able to remove any LLVM dev dependencies?>

@bmulvihill Oh yeah, another advantage of this PR, **LLVM won't be a requirement to compile scry!** :taco: 

```
➜  scry git:(fa/lightweight-scry-speedup) ✗ sudo pacman -Rns llvm
[sudo] password for main: 
checking dependencies...

Packages (1) llvm-5.0.1-2

Total Removed Size:  159.96 MiB

:: Do you want to remove these packages? [Y/n] Y
:: Processing package changes...
(1/1) removing llvm               [#####################################################] 100%
:: Running post-transaction hooks...
(1/1) Arming ConditionNeedsUpdate...
➜  scry git:(fa/lightweight-scry-speedup) ✗ time crystal build src/scry.cr -s -p --no-debug
Parse:                             00:00:00.000311398 (   0.25MB)
Semantic (top level):              00:00:00.379528606 (  52.33MB)
Semantic (new):                    00:00:00.002840074 (  52.33MB)
Semantic (type declarations):      00:00:00.034088660 (  52.33MB)
Semantic (abstract def check):     00:00:00.002330909 (  52.33MB)
Semantic (ivars initializers):     00:00:00.001950688 (  52.33MB)
Semantic (cvars initializers):     00:00:00.250472900 (  68.33MB)
Semantic (main):                   00:00:01.612165740 ( 196.77MB)
Semantic (cleanup):                00:00:00.006179480 ( 196.77MB)
Semantic (recursive struct check): 00:00:00.001546292 ( 196.77MB)
Codegen (crystal):                 00:00:01.210651356 ( 228.77MB)
Codegen (bc+obj):                  00:00:00.387764860 ( 228.77MB)
Codegen (linking):                 00:00:00.714893504 ( 228.77MB)
                                          
Codegen (bc+obj):
 - all previous .o files were reused
crystal build src/scry.cr -s -p --no-debug  4.80s user 0.47s system 112% cpu 4.674 total
➜  scry git:(fa/lightweight-scry-speedup) ✗ rm -rf ~/.cache/crystal                        
➜  scry git:(fa/lightweight-scry-speedup) ✗ time crystal build src/scry.cr -s -p --no-debug
Parse:                             00:00:00.000323728 (   0.25MB)
Semantic (top level):              00:00:00.395329747 (  52.33MB)
Semantic (new):                    00:00:00.002867334 (  52.33MB)
Semantic (type declarations):      00:00:00.036601810 (  52.33MB)
Semantic (abstract def check):     00:00:00.002567500 (  52.33MB)
Semantic (ivars initializers):     00:00:00.002118485 (  52.33MB)
Semantic (cvars initializers):     00:00:00.246402066 (  68.33MB)
Semantic (main):                   00:00:01.628563404 ( 196.77MB)
Semantic (cleanup):                00:00:00.005707169 ( 196.77MB)
Semantic (recursive struct check): 00:00:00.001652108 ( 196.77MB)
Codegen (crystal):                 00:00:01.205903900 ( 228.77MB)
Codegen (bc+obj):                  00:00:06.013242583 ( 228.77MB)
Codegen (linking):                 00:00:00.712104028 ( 228.77MB)
                                          
Codegen (bc+obj):
 - no previous .o files were reused
crystal build src/scry.cr -s -p --no-debug  14.66s user 0.71s system 148% cpu 10.323 total
```

:smile:

faustinoaq:
> How will this affect CI? We should be able to remove any LLVM dev dependencies?

Well, now CI is 3X faster :rocket: 

Before: 5.13s

screenshot_20180305_141434

Now 1.44s :100: 

screenshot_20180305_141333


Memory usage has changed as well, 1GB vs 250MB :sweat_smile:

keplersj:
@bmulvihill I second your comments. I've been swayed. This is a good idea. If there were a way to interact with the compiler programmatically without all the weight I'd be for that, but this works well in this form.

faustinoaq:
So, I think this is ready, WDYT about merging this?

keplersj:
:shipit:

PR #54, opened 2018-03-04
Allow formatting unsaved document

faustinoaq:
Currently scry can't format an unsaved document, with this PR now you can format even a untitled file :tada: 

ezgif-2-db8399697c

/cc @crystal-lang-tools/scry

### DIFF
--- a/src/scry/context.cr
+++ b/src/scry/context.cr
@@ -69,12 +69,14 @@ module Scry
     private def dispatchRequest(params : DocumentFormattingParams, msg)
       text_document = TextDocument.new(params, msg.id)
 
-      unless text_document.untitled?
-        formatter = Formatter.new(@workspace, text_document)
-        response = formatter.run
-        Log.logger.debug(response)
-        response
+      if open_file = @workspace.open_files.first_value?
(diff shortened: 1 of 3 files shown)


laginha87:
Tried it out on untitled and unsaved files and worked well. Takes a bit to be able to use after opening the document. After that initial slowdown it works pretty fast.

faustinoaq:
@laginha87 Thank you for testing it! :+1: 

> After that initial slowdown it works pretty fast.

That slowdown is because scry block the requests using analyzer to check diagnostics in your code via `crystal build` , semantic diagnostics feature and implementation search are pretty slow features. I tried to make it async but seems is harder that I thought https://github.com/crystal-lang-tools/scry/pull/48#issuecomment-370222717

You can repeat the slowdown if you save the file or go to a definition and then try immediately formattting or symbol listing or completion, you'll see that request get blocked

faustinoaq:
@keplersj @bmulvihill Can you review this as well? :sweat_smile:

faustinoaq:
@bmulvihill Thank you for your review, I'm going to merge it :+1:

PR #57, opened 2018-03-08
On ci check that code is formatted

laginha87:


faustinoaq:
I approve this but What about using ameba by @veelenga?

I added it to amber and works pretty well :sweat_smile: 

https://github.com/amberframework/amber/issues/465


Maybe we can add it in other PR, WDYT?

laginha87:
Ameba seems like a good idea as well and we can add it later.
I created this PR because I noticed I had committed unformatted code on the last PR and it was easily avoidable on travis.

faustinoaq:
> Ameba seems like a good idea as well and we can add it later.

Yeah, I opened https://github.com/crystal-lang-tools/scry/issues/58 to track this :+1:

PR #63, opened 2018-03-11
Allow not found implementation

faustinoaq:
Currently if an implementation isn't found the editor keeps waiting something... forever...

_See the blue line_

screenshot_20180311_084649

This change allow to send an empty list of location, so we will see an nice message saying that our implementation/definition isn't found:

screenshot_20180311_085000

This PR also simplifies our diagnostic message by removing `[crystal]` from source title:

Before:

screenshot_20180311_085221

Now:

screenshot_20180311_085307

### DIFF
--- a/src/scry/protocol/diagnostic.cr
+++ b/src/scry/protocol/diagnostic.cr
@@ -43,7 +43,7 @@ module Scry
         Position.new(line - 1, column + size - 1)
       )
       @severity = DiagnosticSeverity::Error.value
-      @source = "Scry [Crystal]"
+      @source = "Scry"
     end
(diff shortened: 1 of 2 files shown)


src/scry/protocol/diagnostic.cr
```diff
@@ -43,7 +43,7 @@ module Scry
         Position.new(line - 1, column + size - 1)
       )
       @severity = DiagnosticSeverity::Error.value
-      @source = "Scry [Crystal]"
+      @source = "Scry"
```
keplersj: Much Appreciated.

Issue #69, opened 2018-03-29
Correctly Handle Shutdown Request


keplersj:
The current handling of this request is not in line with the LSP spec, and is causing issue with Atom editors.

Spec Documentation: https://microsoft.github.io/language-server-protocol/specification#shutdown

Ref: https://github.com/crystal-lang-tools/atom-ide-crystal/issues/32

PR #70, opened 2018-03-29
Handle exit as 'exit' notification not 'shutdown' request

damieng:
Requests should always respond, notifications do not need to.

With that in mind I think 'exit' should do what shutdown currently does in the NotificationMessage hander.

Shutdown still needs to be handled but I think right now it can just return a response and not do anything.  (Not sure how to do that here)

Partial fix for #69


faustinoaq:
Already fixed by #76

PR #76, opened 2018-04-09
Handle exit and shutdown request

faustinoaq:
Fixes https://github.com/crystal-lang-tools/scry/issues/69 and replaces #70 

/cc @damieng

### DIFF
--- a/src/scry/context.cr
+++ b/src/scry/context.cr
@@ -26,8 +26,6 @@ module Scry
 
     # A request message to describe a request between the client and the server.
     # Every processed request must send a response back to the sender of the request.
-    # See, https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#shutdown-request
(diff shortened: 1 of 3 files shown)


faustinoaq:
#75 already cover an spec for this :+1:

src/scry/context.cr
```diff
@@ -27,26 +29,32 @@ module Scry
...
       Log.logger.debug(msg.method)
-      dispatchNotification(msg.params, msg)
+      dispatch_notification(msg.params, msg)
+    end
+
+    private def dispatch_request(params : Nil, msg)
+      if msg.method == "shutdown"
+        Scry.shutdown = true
+        ResponseMessage.new(msg.id, nil)
+      elsif msg.method == "exit"
```
damieng: I think the 'exit' handling needs to be in dispatch_notification as it is a notification message not a request message (unlike shutdown which is in the correct place)

faustinoaq: Oh, yeah, Thank you!

damieng:
Looks good to me.

PR #78, opened 2018-04-09
Search implementations in files with implicit requires

### LINKED ISSUE #29: Support mainFile setting
Some projects could have a main file, avoiding `require`'s in every file. So instead of analyzing each file, with `mainFile` option we can analyze just one. [0] By example, on amber projects [1]

[0] - https://github.com/crystal-lang-tools/vscode-crystal-lang/wiki/Settings#mainfile
[1] - https://github.com/amberframework/amberframework.org/blob/master/src/amberframework.cr#L5

faustinoaq:
**NOTE:** This feature doesn't affect scry performance, just increase the scope available to search methods and dependencies without requiring files explicitly (common practice in crystal projects)

Fixes #29

* Use root_uri from workspace
* Add get_scope to set scope to src folder on standard crystal projects
* Also get_scope sets the scope to the worspace if no src folder exists
* Send empty implementation message if an error is found...
* ...enabling the message "not found implementation" in the LSP client

The rules are:

1. If your project has a `src` directory (common on crystal projects) then you should have a main file that requires all others using `require "./**/*"` or similar
2. If your project doesn't have a `src` folder then all the workspace is analyzed in aims to search symbols and methods, excluding `spec` and `lib` directories from primary scope (methods inside `lib` dir can be found if a primary scope file requires it)

The current spec already cover part of this feature, a full spec for this is a bit complicated because it requires a full workspace to test implicit requires

This PR also include a replacement and enhancement to the fix implemented on #63

vokoscreen-2018-04-09_02-50-12 1

### DIFF
--- a/spec/scry/implementations_spec.cr
+++ b/spec/scry/implementations_spec.cr
@@ -5,7 +5,8 @@ module Scry
     it "check implementation response type" do
       params = TextDocumentPositionParams.from_json(TEXTDOCUMENT_POSITION_PARAM_EXAMPLE)
       text_document = TextDocument.new(params, 0)
-      impl = Implementations.new(text_document)
+      workspace = Workspace.new("root_uri", 0, 10)
+      impl = Implementations.new(workspace, text_document)
       impl.run.is_a?(ResponseMessage).should eq(true)
     end
   end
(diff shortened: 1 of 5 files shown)


faustinoaq:
BTW, I tested this feature with an amber project and worked out-the-box

vokoscreen-2018-04-09_03-44-21

^ Notice I don't have any `require` at beginning of most of these files
^ Also notice the errors, now the diagnostic system is independent of search implementation, although, if the implementation tool finds a "valid" error (syntax, truly  missing symbols, etc), then it will return "Implementation not found"

PR #80, opened 2018-04-09
Analyze files with implicit requires

### LINKED ISSUE #29: Support mainFile setting

faustinoaq:
This PR basically adds the following feature: Analyze the code twice if undefined symbol is found on diagnostics result

The rules:

1. first time analyzes a single file and stop if no error
2. second time analyzes the entire project requiring the first level files inside src folder

**Summary of changes:**

* Add method inside_crystal_path? to avoid analizing lib and stdlib
* Send clean diagnostics for all files when a full project is analyzed
* Allow analyze files with implicit requires (main file style)
* Adds --no-debug flag for speed up compilation
* Add fake file .scry.cr to analyze a full project (aka main file)
* Remove the fake file .scry.cr from diagnostic list
* Add untitled: to the uri_to_filename method

**Use case:** very useful on those projects where you have a main file that require all other files, allowing you to write your code without multiple `require`s in every file

Fixes #29 as well

Related to #78 (This allows to remove the errors across files when you search implementations)

### DIFF
--- a/src/scry/analyzer.cr
+++ b/src/scry/analyzer.cr
@@ -9,27 +9,42 @@ module Scry
     end
 
     def run
-      @text_document.text.map do |text|
-        unless inside_path?
+      if @text_document.inside_crystal_path?
+        clean_diagnostic
+      else
+        @text_document.text.map do |text|
           analyze(text)
         end
       end.flatten.uniq
     end
 
-    private def inside_path?
-      ENV["CRYSTAL_PATH"].split(':').each do |path|
-        return true if @text_document.filename.starts_with?(path)
+    # Reset all diagnostics in the current project
+    def clean_diagnostic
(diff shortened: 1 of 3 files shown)


faustinoaq:
**About the spec:**, as same as I said on #78, Actually the current test already cover this feature. To create a full test for this feature we need a full workspace, so is a bit harder that common specs.I think those kind of specs can be implemented on other PR.

PR #94, opened 2018-04-22
Add implementation fixes and parse output properly using JSON.mapping

faustinoaq:
Another fix :tada: 

This changes `JSON.parse` by `JSON.mapping`

Fixes #89 

**Edit:**

Add support for main file `.scry.cr` to customize implementation scope

PR #95, opened 2018-04-23
Support custom main file .scry_main.cr and add some diagnostic fixes

faustinoaq:
This PR depends on #94 (Should be merged only after that)

Related to #80

Currently to support diagnostics with implicit requires, scry is already using a "ghost" file `.scry_main.cr`,  to analyze files inside `./src/*`. although, this use case can change, and people that don't use src folder still can use implicit requires by adding a `.scry_main.cr` file with their `require`s on workspace root

screenshot_20180423_142417

### DIFF
--- a/src/scry/publish_diagnostic.cr
+++ b/src/scry/publish_diagnostic.cr
@@ -6,9 +6,14 @@ require "./workspace"
 
 module Scry
   struct PublishDiagnostic
-    FILES_WITH_DIAGNOSTIC = [] of String
+    # Store all groups of files that have received some diagnostic,
+    # Useful to clean diagnostics after code changes.
+    # Each group depends on the file that generated the diagnostic
+    # and can be cleared just by that file.
+    ALL_FILES_WITH_DIAGNOSTIC = {} of String => Array(String)
 
     def initialize(@workspace : Workspace, @uri : String)
+      ALL_FILES_WITH_DIAGNOSTIC[@uri] = [] of String
     end
 
     private def notification(params)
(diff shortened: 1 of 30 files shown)


faustinoaq:
This PR should be updated with master after #94 is merged

src/scry/publish_diagnostic.cr
```diff
@@ -6,7 +6,8 @@ require "./workspace"
 
 module Scry
   struct PublishDiagnostic
-    METHOD = "textDocument/publishDiagnostics"
+    METHOD                = "textDocument/publishDiagnostics"
+    FILES_WITH_DIAGNOSTIC = [] of String
```
bew: Why it that a constant and not an ivar?

faustinoaq: @bew Oh, I just realize we don't need that, we don't need even a var, this is used just on one place, let me fix that :+1:

faustinoaq: Done!

bew: Huh, I was talking about `FILES_WITH_DIAGNOSTIC` :joy:

src/scry/publish_diagnostic.cr
```diff
@@ -6,20 +6,31 @@ require "./workspace"
 
 module Scry
   struct PublishDiagnostic
-    METHOD = "textDocument/publishDiagnostics"
+    FILES_WITH_DIAGNOSTIC = [] of String
```
bew: huh I was talking about this constant :joy:

faustinoaq: Well, this constant is required, `FILES_WITH_DIAGNOSTIC` is updated only during `@diagnostic.from`, so, if I use an ivar, then next time I fix my code, I won't be able to clean the past errors :sweat_smile: 

I could use a class var as well, but I think a constant is cleaner

src/scry/publish_diagnostic.cr
faustinoaq: See https://github.com/crystal-lang-tools/scry/pull/95#discussion_r183521067

src/scry/publish_diagnostic.cr
bew: Oh ok, could you rename it as `FILES_WITH_DIAGNOSTIC` with a little comment about that?

faustinoaq: @bew Done!, I just discovered this was clearing all diagnostics, even those not generated by a current file, so, I'm using a Hash now to group diagnostic per file:

Imgur

faustinoaq:
#94 is already merged! :tada: waiting for more reviews :+1:

faustinoaq:
@bew What about just `.main.cr` ?

bew:
No, I prefer to have `scry` in the name, so I know why this file is there

PR #96, opened 2018-04-27
Add integration tests for autocomplete

laginha87:
Hi all,
when I added autcomplete I felt that the tests where too implementation specific, this PR adds specs that test the completion features as close to end to end as I could, it sends a jsonrpc message and a parses the json response to check the results. Also wanted an excuse to play around with macros :D. 
This way we're not really tied to the implementation we have, actually ended up fixing an issue when completing propertys, getters and setters.

### DIFF
--- a/spec/scry/completion_provider_spec.cr
+++ b/spec/scry/completion_provider_spec.cr
@@ -32,13 +32,13 @@ module Scry
     _context_ = Context.new
     root_path = File.expand_path("spec/fixtures/completion/")
     _context_.test_send_init(root_path)
-    _file_path_ = File.join(root_path, "tree.cr")
+    _file_path_ = File.join(root_path, "sample.cr")
 
     context "module completion" do
       _kind_ = CompletionItemKind::Module
 
       it_completes("require \"arr", ["array"])
-      it_completes("require \"./sa", ["sample"])
+      it_completes("require \"./tr", ["tree"])
     end
 
     context "method completion" do
@@ -62,6 +62,7 @@ module Scry
         it_completes("a = true
                       a.to_", %w(to_json to_s to_s to_yaml))
       end
+
       context "int32" do
         _expected_labels_ = INT32_METHODS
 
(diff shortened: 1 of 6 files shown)


spec/scry/completion_provider_spec.cr
```diff
@@ -1,38 +1,133 @@
 require "../spec_helper"
 
 module Scry
+  BOOLEAN_METHODS = %w(!= & == ^ clone hash to_json to_s to_s to_yaml |)
+  INT32_METHODS   = %w(- clone popcount)
+
+  private macro it_completes(code, with_labels = nil, file = __FILE__, line = __LINE__)
+
+    it "completes #{{{code}}.dump}" do
```
bew: I think you forgot `, {{file}}, {{line}}` after the description

spec/scry/completion_provider_spec.cr
```diff
@@ -1,38 +1,133 @@
...
+      {% if with_labels %}
+        %expected = {{with_labels}}
+      {% else %}
+        %expected = _expected_labels_
+      {% end %}
+
+      %code = {{code}}
+      cursor_location = [%code.lines.size, %code.size]
```
bew: You'd better use a tuple here (no need to create a full array for that), so `{%code.lines.size, %code.size}`

laginha87: I tried to do this initially but was getting a syntax error I didn't understand.
Went back and tried it again and noticed the issue was that `{%code` would be interpreted as macro expression or something, had to add a space between `{` and `%` for it to work :( 
Thanks for the review 👍

spec/scry/completion_provider_spec.cr
bew: oh right, didn't thought about that :+1:

spec/scry/completion_provider_spec.cr
```diff
@@ -14,7 +14,7 @@ module Scry
       {% end %}
 
       %code = {{code}}
-      cursor_location = [%code.lines.size, %code.size]
+      cursor_location = { %code.lines.size, %code.size }
```
bew: the formatter will remove the spaces here :smiley:

faustinoaq: @bew I think `crystal tool format` doesn't work inside macros :sweat_smile:

bew: I have a formatter embedded in my brain, and it removes the spaces here :stuck_out_tongue:

laginha87:
@crystal-lang-tools/scry  any chance I could get an approval here? This would be super helpful for fixing some other completion issues.

faustinoaq:
Hey @crystal-lang-tools/scry Can we get another approval here? :sweat_smile:

PR #92, opened 2018-04-22
Check file inside dependency_graph before using it

faustinoaq:
Hi scry community! :grinning: 

More fixes :tada: 

This check availability for a file inside `dependency_graph`  before call it. 

Useful for untitled files and external files

Fixes #91 
Fixes #59 

/cc @laginha87

PR #98, opened 2018-04-30
Reload completion on file changes

### LINKED ISSUE #61: Completion: Dependancy graph and method db should be updated when there import changes
if you have this scenario.
```crystal
# File b.cr
class A
   def self.a
   end
end
```

```crystal
# File a.cr
```

and then change file a.cr to:

```crystal
# File a.cr
require "./b"
A.
```

Completion on A. doesn't work. 
This is because the dependancy to b.cr wasn't added to the file.
We can either change this on save or on update. But I think on update might be really slow.

faustinoaq:
This PR does:

- ~~For some reason the graph is not getting updated yet~~ Fixed! :tada:  see https://github.com/crystal-lang-tools/scry/pull/98#issuecomment-385419208
- Based on #92 (This PR depends on #92)

### Changes

- Reload graph for new file changes given a `text_document.filename`
- Delete completion file node from dependency graph on drop_file
- Fixes #61 

/cc @laginha87

### DIFF
--- a/src/scry/workspace.cr
+++ b/src/scry/workspace.cr
@@ -38,7 +38,8 @@ module Scry
       else
         file_dependencies = @dependency_graph.prelude_node.descendants.map &.value
       end
-      method_db = Completion::MethodDB.generate(file_dependencies)
+      file_dependencies << file.filename
+      method_db = Completion::MethodDB.generate(file_dependencies.uniq!)
       @open_files[file.filename] = {file, method_db}
     end
 
--- a/src/scry/workspace.cr
+++ b/src/scry/workspace.cr
@@ -24,8 +24,10 @@ module Scry
       @dependency_graph = Completion::DependencyGraph::Builder.new(@lookup_path).build
     end
 
-    def reopen_workspace(filename)
-      @dependency_graph = Completion::DependencyGraph::Builder.new(@lookup_path).rebuild(@dependency_graph, filename)
+    def reopen_workspace(file)
+      @dependency_graph = Completion::DependencyGraph::Builder.new(@lookup_path).rebuild(@dependency_graph, file.filename)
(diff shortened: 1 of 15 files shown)


laginha87:
Damn was also working on fixing this one :P 
Looks like the right way to go 👍

faustinoaq:
@laginha87 Yeah, I;m trying to fix #61, although, this PR doesn't work yet. The graph is not updated on `.rebuild` and I don't know why :sweat_smile: 

Maybe I'm checking the wrong file? on `if file == filename` ?

laginha87:
that bit is a bit overkill you should be able to just run `process_requires` on the filename
Also you need to update the method_db for the open file for it to work. For now a quick test you can do for the dependancy graph part is close and open the file.

faustinoaq:
@crystal-lang-tools/scry Can I get another approval here? :sweat_smile:

src/scry/workspace.cr
```diff
@@ -9,16 +9,26 @@ module Scry
...
 
+    @lookup_path : Array(String)
+
     def initialize(root_uri, process_id, max_number_of_problems)
       @root_uri = root_uri
       @process_id = process_id
       @max_number_of_problems = max_number_of_problems
       @open_files = {} of String => {TextDocument, Completion::MethodDB}
       @dependency_graph = Completion::DependencyGraph::Graph.new
+      @lookup_path = ENV["CRYSTAL_PATH"].split(":") + [@root_uri]
```
bew: Small note: Remember: If you're doing `+ [expr]` you're having 2 new arrays!

faustinoaq: What about `ENV["CRYSTAL_PATH"].split(":") << @root_uri` ?

PR #100, opened 2018-05-04
Add class and module name completion

laginha87:
Hi all
This adds autocomplete for class names and module names.
It doesn't show any documentation for it yet.
Will add a gif to demo.

### DIFF
--- a/spec/fixtures/completion/sample.cr
+++ b/spec/fixtures/completion/sample.cr
@@ -1,6 +1,6 @@
 require "./tree"
 
-class A
+class B
   def method
   end
 end
--- a/spec/scry/completion_provider_spec.cr
+++ b/spec/scry/completion_provider_spec.cr
@@ -127,6 +127,9 @@ module Scry
                           a."
 
         it_completes "Node.", %w(new)
+
+        it_completes "a = A.new
+                      a.", %w(method)
       end
     end
   end
--- a/spec/scry/completion_provider_spec.cr
+++ b/spec/scry/completion_provider_spec.cr
@@ -136,8 +136,11 @@ module Scry
     context "classname completion" do
       _kind_ = CompletionItemKind::Class
       it_completes("A", %w(Array Atomic ArgumentError AtExitHandlers))
+      it_completes(" ::A", %w(Array Atomic ArgumentError AtExitHandlers))
+      it_completes(" ::", %w())
       it_completes("JSON::P", %w(ParseException Parser PullParser))
       it_completes("JSON::Pa", %w(ParseException Parser))
+      it_completes("JSON::", %w(Any Builder Error Lexer Lexer::IOBased Lexer::StringBased ParseException Parser PullParser Token))
       it_completes("JSO", %w(JSON JSON::Any JSON::Builder JSON::Error JSON::Lexer JSON::Lexer::IOBased JSON::Lexer::StringBased JSON::ParseException JSON::Parser JSON::PullParser JSON::Token))
     end
   end
--- a/spec/scry/completion_provider_spec.cr
+++ b/spec/scry/completion_provider_spec.cr
@@ -128,16 +128,19 @@ module Scry
 
         it_completes "Node.", %w(new)
 
-        it_completes "a = A.new
+        it_completes "a = B.new
                       a.", %w(method)
       end
     end
 
     context "classname completion" do
       _kind_ = CompletionItemKind::Class
-      it_completes("A", %w(Array Atomic ArgumentError))
+      it_completes("A", %w(Array Atomic ArgumentError AtExitHandlers))
       it_completes("JSON::P", %w(ParseException Parser PullParser))
       it_completes("JSON::Pa", %w(ParseException Parser))
+      it_completes("JSO", %w(JSON JSON::Any JSON::Builder JSON::Error JSON::Lexer JSON::Lexer::IOBased JSON::Lexer::StringBased JSON::ParseException JSON::Parser JSON::PullParser JSON::Token))
     end
   end
 end
+
+
--- a/spec/scry/completion_provider_spec.cr
+++ b/spec/scry/completion_provider_spec.cr
@@ -142,5 +142,3 @@ module Scry
     end
   end
 end
-
-
--- a/spec/scry/completion_provider_spec.cr
+++ b/spec/scry/completion_provider_spec.cr
@@ -133,7 +133,7 @@ module Scry
       end
     end
 
-    context "classname completion" do
+    context "module, struct and class name completion" do
       _kind_ = CompletionItemKind::Class
       it_completes("A", %w(Array Atomic ArgumentError AtExitHandlers))
       it_completes(" ::A", %w(Array Atomic ArgumentError AtExitHandlers))
--- a/spec/scry/completion_provider_spec.cr
+++ b/spec/scry/completion_provider_spec.cr
@@ -132,5 +132,12 @@ module Scry
                       a.", %w(method)
       end
     end
+
+    context "classname completion" do
+      _kind_ = CompletionItemKind::Class
+      it_completes("A", %w(Array Atomic ArgumentError))
+      it_completes("JSON::P", %w(ParseException Parser PullParser))
+      it_completes("JSON::Pa", %w(ParseException Parser))
+    end
   end
 end
--- a/src/scry/completion/class_module_context.cr
+++ b/src/scry/completion/class_module_context.cr
@@ -8,7 +8,10 @@ module Scry::Completion
     def find
       target_header_index = @target.rindex(":")
       @method_db.type_match(@target).map do |label|
-        label = label[(target_header_index + 1)..-1] if target_header_index
+        if target_header_index
+          label = label[(target_header_index + 1)..-1]
+        end
+
         CompletionItem.new(
           label: label,
           insert_text: label,
--- a/src/scry/completion/class_module_context.cr
+++ b/src/scry/completion/class_module_context.cr
@@ -1,0 +1,22 @@
+require "./context"
+
+module Scry::Completion
+  class ClassModuleContext < Context
+    def initialize(@text : String, @target : String, @method_db : Completion::MethodDB)
+    end
+
+    def find
+      target_header_index = @target.rindex(":")
+      @method_db.type_match(@target).map do |label|
+        label = label[(target_header_index + 1)..-1] if target_header_index
+        CompletionItem.new(
+          label: label,
+          insert_text: label,
+          kind: CompletionItemKind::Class,
+          detail: label,
+          data: nil
+        )
+      end
+    end
+  end
+end
--- a/src/scry/completion/method_db.cr
+++ b/src/scry/completion/method_db.cr
@@ -20,6 +20,10 @@ module Scry::Completion
       end.select(&.name.starts_with? text).to_a
     end
 
+    def type_match(text : String) : Array(String)
+      @db.keys.select(&.starts_with?(text)).reject(&.ends_with?(".class"))
+    end
+
     def self.generate(paths)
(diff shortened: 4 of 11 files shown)


faustinoaq:
image

Works like a charm! :tada:

spec/scry/completion_provider_spec.cr
```diff
@@ -128,9 +128,17 @@ module Scry
...
                       a.", %w(method)
       end
     end
+
+    context "classname completion" do
+      _kind_ = CompletionItemKind::Class
+      it_completes("A", %w(Array Atomic ArgumentError AtExitHandlers))
+      it_completes("JSON::P", %w(ParseException Parser PullParser))
+      it_completes("JSON::Pa", %w(ParseException Parser))
+      it_completes("JSO", %w(JSON JSON::Any JSON::Builder JSON::Error JSON::Lexer JSON::Lexer::IOBased JSON::Lexer::StringBased JSON::ParseException JSON::Parser JSON::PullParser JSON::Token))
```
bew: Can you add a spec for completion after `JSON::` ?
And after `::` ? (if it works)

src/scry/completion/method_db.cr
```diff
@@ -20,6 +20,10 @@ module Scry::Completion
       end.select(&.name.starts_with? text).to_a
     end
 
+    def type_match(text : String) : Array(String)
+      @db.keys.select(&.starts_with?(text)).reject(&.ends_with?(".class"))
```
bew: `select!`
`reject!`

src/scry/completion/method_db.cr
```diff
@@ -53,39 +57,54 @@ module Scry::Completion
 
   class Visitor < Crystal::Visitor
     property classes
+    property class_name = ""
```
bew: I think it's better to use `nil` to denote the absence of `class_name`, so the compiler can help you instead of doing empty checks.

src/scry/completion/method_db.cr
```diff
@@ -53,39 +57,54 @@ module Scry::Completion
...
       @classes[method_receiver] << MethodDBEntry.new(name, signature, @file, node.location.to_s)
       false
     end
 
     def visit(node : Crystal::ClassDef)
-      @classes["#{node.name.to_s}.class"] = [] of MethodDBEntry
-      @classes[node.name.to_s] = [] of MethodDBEntry
-      @class_queue << node.name.to_s
+      @class_queue << node.name
+      @class_name = generate_name(@module_queue + @class_queue)
```
bew: Does this still work if you have a module inside a class inside a module?

Why are you separating module & classes queues?

src/scry/completion/method_db.cr
faustinoaq: @bew Why? are mutable methods faster?

bew: Yeah I think so, they don't allocate a new array each time.

src/scry/completion/method_db.cr
laginha87: oh, I didn't even know that could happen

laginha87: It's seperate cause we only store methods for class, for now.

src/scry/completion/method_db.cr
laginha87: I think it's easier with an empty string cause we only interpolate stuff, otherwise I'll have to use not_nil! calls

laginha87: @class_name is assigned  the value of the generate_classname which will be an empty string if the queues are empty.

faustinoaq:
```crystal
property class_name : String?
```

and

```crystal
if this_class_name = class_name
  # Use this_cass_name here...
end
```

WDYT?

faustinoaq: Although I see no issues with initializing the `class_name = ""` 

@bew Is this a bad practice or something? :sweat_smile:

bew: I'll simply say:

```
method_receiver = node.receiver ? "#{@class_name}.class" : @class_name
```
In this case, does it make sense (allowed) to have `@class_name` as an empty string?
I don't think so, so it must be made difficult/impossible to do.

laginha87: that won't happen cause if class_name is "" then class_queue is empty and in here we opt out https://github.com/crystal-lang-tools/scry/pull/100/files/d05348c07392320c379be63575985d0f83398c56#diff-45d877a34e1040cf2af4c4c46cdcf290R73
The idea of the classname is just to cache the output of generate_class_name which never returns nil, so initialising it as nil doesn't seem like the right way to handle it.

faustinoaq: @laginha87 Oh, ok, I think `class_name = ""` is not that bad @bew WDYT?

bew: yeah let's not block the PR for a tiny non-straight-forward thing like that, it's ok

spec/scry/completion_provider_spec.cr
```diff
@@ -136,8 +136,11 @@ module Scry
     context "classname completion" do
       _kind_ = CompletionItemKind::Class
       it_completes("A", %w(Array Atomic ArgumentError AtExitHandlers))
+      it_completes(" ::A", %w(Array Atomic ArgumentError AtExitHandlers))
+      it_completes(" ::", %w())
```
bew: why nothing? It should give every top level classes & constants

faustinoaq: @bew I think listing all classes would be very slow :sweat_smile: , no?

By example on Workspace symbols, if query is empty, you get this:

screenshot_20180509_193607

bew: @faustinoaq you're showing symbols here, not completion.

Slow? I don't know. 

So for now it only gives an empty list?

laginha87: I couldn't figure out why it didn't return anything :(

src/scry/completion/method_db.cr
laginha87: @bew I tried to change this to use classes and modules in the same queue but started getting some tests failing not really sure why. I cut a new branch to add this and module functions as well.

spec/scry/completion_provider_spec.cr
faustinoaq: No problem, I think is ok :+1:

bew: @laginha87 let's not block this pr as it's not so important, but please open an issue to keep track of `::` completion, and fix it one day, or decide what we do about it

laginha87: I actually wanna tackle doing a context aware comlpetion for this at some point .
for example:
```
module A
     ::
```
should complete with modules and classes inside A as well

laginha87: @bew added https://github.com/crystal-lang-tools/scry/issues/118

bew: @laginha87 there is no need for contextual completion, `::` completions will ALWAYS be the top level things, that's what it exists for, there is no contextual things going on for types that starts with `::`

faustinoaq:
@crystal-lang-tools/scry Can we get another review here?

laginha87:
@bew, thanks for reviewing the code, can you approve this or is there something I missed?

src/scry/completion/method_db.cr
bew: can you create a new issue to track and not forget this? It's quite important
thanks

bew: Also, include what you tried and your results, it might help find the actual problem ;)

spec/scry/completion_provider_spec.cr
```diff
@@ -128,9 +128,20 @@ module Scry
 
         it_completes "Node.", %w(new)
 
-        it_completes "a = A.new
+        it_completes "a = B.new
                       a.", %w(method)
       end
     end
+
+    context "classname completion" do
```
bew: Note: it can also complete module names (and structs), not only class names, maybe change the spec name?

spec/scry/completion_provider_spec.cr
```diff
@@ -128,9 +128,20 @@ module Scry
...
+      _kind_ = CompletionItemKind::Class
+      it_completes("A", %w(Array Atomic ArgumentError AtExitHandlers))
```
bew: This means that this spec rely on external code... A change to the stdlib, an addition/removal of a class starting with `A` will break this spec...

There are other problems with how the specs currently works, like:
- You can't add a description of what that spec is actually spec-ing
- They rely on external code (see above)
- They are not self-contained: by just reading the spec you don't directly have all the data to understand it (you have to always lookup other files, AND they are shared between many specs)

This should probably go to a new issue though. _I'm just too tired right now_ ^^

src/scry/completion/class_module_context.cr
```diff
@@ -0,0 +1,22 @@
...
+
+module Scry::Completion
+  class ClassModuleContext < Context
+    def initialize(@text : String, @target : String, @method_db : Completion::MethodDB)
+    end
+
+    def find
+      target_header_index = @target.rindex(":")
+      @method_db.type_match(@target).map do |label|
+        label = label[(target_header_index + 1)..-1] if target_header_index
```
bew: just a style comment: At first I didn't saw the `if` at the end of the line, and was wondering 'how the hell does that compile..."

I think that line is too 'complicated' / 'long' for a suffix `if`, and would suggest to make it a multi-line `if`.

src/scry/completion/class_module_context.cr
```diff
@@ -0,0 +1,22 @@
...
+        CompletionItem.new(
+          label: label,
+          insert_text: label,
+          kind: CompletionItemKind::Class,
+          detail: label,
```
bew: would it be better to have the full path (e.g: `JSON::Any`, not just `Any`) in details? (can someone test this, see how it goes?)

src/scry/completion/method_db.cr
```diff
@@ -20,6 +20,10 @@ module Scry::Completion
       end.select(&.name.starts_with? text).to_a
     end
 
+    def type_match(text : String) : Array(String)
+      @db.keys.select!(&.starts_with?(text)).reject!(&.ends_with?(".class"))
```
bew:
```cr
@db.keys.select! { |k| k.starts_with?(text) && !k.ends_with?(".class") }
```
? (less allocations / simpler logic / ...)

src/scry/completion/class_module_context.cr
laginha87: you mean in the CompletionItem class ?

spec/scry/completion_provider_spec.cr
laginha87: 1. The test will always break with changes to stdlib, we are testing the autocomplete on it I don't think it's something we should worry about.
2. What other files do you have to check to understand the specs? My idea for this file was to have tests that almost did an end-to-end test on the completion features and not dependant on the actual implementation.

laginha87:
@bew thanks for the review 👍 
I addressed most of the issues you pointed out.
What you brought up about the test is valid but think it should be taken care of in another PR.
Didn't understand what you meant by "would it be better to have the full path (e.g: JSON::Any, not just Any) in details? (can someone test this, see how it goes?)" 
Can you take another look and maybe approve 😅?

bew:
Actually it's not important

spec/fixtures/completion/sample.cr
```diff
@@ -1,6 +1,6 @@
 require "./tree"
 
-class A
+class B
```
bew: By "looking another file to understand the specs" I meant this file!

faustinoaq:
@laginha87 Thank you! :tada:

PR #106, opened 2018-05-05
Add workspace symbols

### LINKED ISSUE #102: Support workspace symbols
Hi community!

@bmulvihill already implemented listing symbols per file via #39 and #43 

Would be nice to use the new "main file" `.scry.cr` to get workspace symbols, WDYT?

`.scry.cr` file is optional, implemented on #95

faustinoaq:
Fixes #102 :tada: 

screenshot_20180505_122258

### DIFF
--- a/spec/scry/symbol_spec.cr
+++ b/spec/scry/symbol_spec.cr
@@ -98,5 +98,21 @@ module Scry
         result.kind.is_a?(SymbolKind::Constant).should be_true
       end
     end
+
+    describe "WorkspaceSymbols" do
+      it "return empty Symbols list if no query" do
+        processor = WorkspaceSymbolProcessor.new(0, ROOT_PATH, "")
+        response = processor.run
+        result = response.result.as(Array(SymbolInformation))
+        result.empty?.should be_true
+      end
+
+      it "return Symbols list with query match" do
+        processor = WorkspaceSymbolProcessor.new(0, "#{ROOT_PATH}/src", "salut")
+        response = processor.run
+        result = response.result.as(Array(SymbolInformation)).try(&.first)
+        result.kind.is_a?(SymbolKind::Method).should be_true
+      end
+    end
   end
 end
--- a/spec/scry/symbol_spec.cr
+++ b/spec/scry/symbol_spec.cr
@@ -128,11 +128,11 @@ module Scry
         result.kind.is_a?(SymbolKind::Class).should be_true
       end
 
-      it "return stdlib symbol with regex query match for puts" do
-        processor = WorkspaceSymbolProcessor.new(0, "#{ROOT_PATH}/src", "put*")
+      it "return stdlib symbol with regex query match for initialize" do
+        processor = WorkspaceSymbolProcessor.new(0, "#{ROOT_PATH}/src", "initializ*")
         response = processor.run
         result = response.result.as(Array(SymbolInformation)).first
-        result.kind.is_a?(SymbolKind::Function).should be_true
+        result.kind.is_a?(SymbolKind::Method).should be_true
       end
     end
   end
--- a/spec/scry/symbol_spec.cr
+++ b/spec/scry/symbol_spec.cr
@@ -19,7 +19,7 @@ module Scry
       text_document = TextDocument.new("uri", ["class Test; end"])
       processor = SymbolProcessor.new(text_document)
       response = processor.run
-      result = response.result.as(Array(SymbolInformation)).try(&.first)
+      result = response.result.as(Array(SymbolInformation)).first
       result.kind.is_a?(SymbolKind::Class).should be_true
     end
 
@@ -27,7 +27,7 @@ module Scry
       text_document = TextDocument.new("uri", ["struct Test; end"])
       processor = SymbolProcessor.new(text_document)
       response = processor.run
-      result = response.result.as(Array(SymbolInformation)).try(&.first)
+      result = response.result.as(Array(SymbolInformation)).first
       result.kind.is_a?(SymbolKind::Class).should be_true
     end
 
@@ -35,7 +35,7 @@ module Scry
       text_document = TextDocument.new("uri", ["module Test; end"])
       processor = SymbolProcessor.new(text_document)
       response = processor.run
-      result = response.result.as(Array(SymbolInformation)).try(&.first)
+      result = response.result.as(Array(SymbolInformation)).first
       result.kind.is_a?(SymbolKind::Module).should be_true
     end
 
@@ -43,7 +43,7 @@ module Scry
       text_document = TextDocument.new("uri", ["def test; end"])
       processor = SymbolProcessor.new(text_document)
       response = processor.run
-      result = response.result.as(Array(SymbolInformation)).try(&.first)
+      result = response.result.as(Array(SymbolInformation)).first
       result.kind.is_a?(SymbolKind::Method).should be_true
     end
 
@@ -51,7 +51,7 @@ module Scry
       text_document = TextDocument.new("uri", ["@bar = nil"])
       processor = SymbolProcessor.new(text_document)
       response = processor.run
-      result = response.result.as(Array(SymbolInformation)).try(&.first)
+      result = response.result.as(Array(SymbolInformation)).first
       result.as(SymbolInformation).kind.is_a?(SymbolKind::Variable).should be_true
     end
 
@@ -86,7 +86,7 @@ module Scry
         text_document = TextDocument.new("uri", [%(HELLO = "world")])
         processor = SymbolProcessor.new(text_document)
         response = processor.run
-        result = response.result.as(Array(SymbolInformation)).try(&.first)
+        result = response.result.as(Array(SymbolInformation)).first
         result.kind.is_a?(SymbolKind::Constant).should be_true
       end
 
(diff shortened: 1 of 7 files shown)


src/scry/symbol.cr
```diff
@@ -115,4 +116,26 @@ module Scry
       ResponseMessage.new(@text_document.id, visitor.symbols)
     end
   end
+
+  class WorkspaceSymbolProcessor
+    def initialize(@msg_id : Int32 | String, @root_path : String, @query : String)
+      @all_files = Dir.glob("#{root_path}/**/*.cr")
```
bew: Why doesn't this use the workspace, and find files based on prelude and requires?

spec/scry/symbol_spec.cr
```diff
@@ -98,5 +98,21 @@ module Scry
...
+        processor = WorkspaceSymbolProcessor.new(0, ROOT_PATH, "")
+        response = processor.run
+        result = response.result.as(Array(SymbolInformation))
+        result.empty?.should be_true
+      end
+
+      it "return Symbols list with query match" do
+        processor = WorkspaceSymbolProcessor.new(0, "#{ROOT_PATH}/src", "salut")
+        response = processor.run
+        result = response.result.as(Array(SymbolInformation)).try(&.first)
```
bew: `try` is uselesw here as you can the object to a non-nilable type juste before.

spec/scry/symbol_spec.cr
```diff
@@ -98,5 +98,21 @@ module Scry
...
+    describe "WorkspaceSymbols" do
+      it "return empty Symbols list if no query" do
+        processor = WorkspaceSymbolProcessor.new(0, ROOT_PATH, "")
+        response = processor.run
+        result = response.result.as(Array(SymbolInformation))
+        result.empty?.should be_true
+      end
+
+      it "return Symbols list with query match" do
+        processor = WorkspaceSymbolProcessor.new(0, "#{ROOT_PATH}/src", "salut")
```
bew: Use regex here? Or add another spec for a regex query?

src/scry/symbol.cr
faustinoaq: > and find files based on prelude and requires?

What do you mean? :sweat_smile: This will get all "workspace" files to analyze all "workspace" symbols

spec/scry/symbol_spec.cr
faustinoaq: Done! :+1:

src/scry/symbol.cr
bew: Oh ok, so this symbol feature is just to search in your project, not everywhere (stdlib, other shards, etc) ?

src/scry/symbol.cr
```diff
@@ -115,4 +116,26 @@ module Scry
...
+    end
+
+    def run
+      symbols = [] of SymbolInformation
+      unless @query.empty?
+        @all_files.each do |file|
+          visitor = SymbolVisitor.new("file://#{file}")
+          parser = Crystal::Parser.new(File.read(file))
+          parser.filename = file
+          node = parser.parse
```
bew: what if there is a syntax error in one of the file? I think the exception needs to be rescued and the file skipped.

Is it possible to send errors in addition to valid results? (say, here are the results for files X & Y, but there is an error with file Z)

src/scry/symbol.cr
bew: Ok it's fine then. I think I'd even restrict to `src`, to not show implementation detail symbols from shards

src/scry/symbol.cr
faustinoaq: I guess I can handle the error, but the response should still be managed by diagnostic provider

I think I can return an empty array if syntax error is found

bew: Looking at https://microsoft.github.io/language-server-protocol/specification#workspace-symbols-request-leftwards_arrow_with_hook you should return null and an error in case an exception happens during the workspace symbol request.

I think it could be possible to return a result & an error? but how the valid_result/error is handled would vastly depend on the language client implementation..

faustinoaq: Well, Actually we're managing that via `response_error.cr`, although I think for better user experience we should ignore parsing errors here since the diagnostics already do this work.

I don't want to see a popup saying "LSP server has an error" every time I try to list symbols and I have files with syntax errors :sweat_smile:

src/scry/symbol.cr
faustinoaq: @bew Well, I guess I like your idea of listing stdlib symbols, so, this feature will support listing symbols from:

1. Your code
2. Your shards
3. Crystal stdlib

screenshot_20180505_181554

And works very fast!!! :tada: and stdlib is memoized :sparkles:

faustinoaq:
screenshot_20180505_181554
screenshot_20180505_182214

stdlib symbols has been added as well :tada:

src/scry/symbol.cr
```diff
@@ -114,28 +114,44 @@ module Scry
...
 
   class WorkspaceSymbolProcessor
+    @@crystal_path_files = [] of SymbolInformation
+
     def initialize(@msg_id : Int32 | String, @root_path : String, @query : String)
-      @all_files = Dir.glob("#{root_path}/**/*.cr")
+      @workspace_files = Dir.glob(File.join(root_path, "**", "*.cr"))
+      if @@crystal_path_files.empty?
+        # Memoize crystal stdlib
+        @@crystal_path_files = search_symbols(Dir.glob(File.join(Scry.default_crystal_path, "**", "*.cr")), ".*")
```
bew: rename to `crystal_path_symbols` as it stores symbols

src/scry/symbol.cr
```diff
@@ -114,28 +114,44 @@ module Scry
...
-        @all_files.each do |file|
-          visitor = SymbolVisitor.new("file://#{file}")
-          parser = Crystal::Parser.new(File.read(file))
-          parser.filename = file
-          node = parser.parse
-          node.accept(visitor)
-          symbols.concat visitor.symbols.select(&.name.match(Regex.new(@query)))
-        end
+        symbols.concat search_symbols(@workspace_files, @query)
+        symbols.concat @@crystal_path_files.select(&.name.match(Regex.new(@query)))
```
bew: You could store the `Regex` object in `@query` in the `initialize`, instead of storing the `String` and re-create/re-compile the regex on each usage.

src/scry/symbol.cr
faustinoaq: Done :+1:

bew:
Also, why in you last screenshot some entries have non-absolute path even though it's not in the current workspace (e.g: `pow_f64` or `FileModule`) ?

faustinoaq:
@bew Interesting I didn't realize of that :+1:

I think `SymbolInformation` is too long and VSCode just hide it :sweat_smile: 

If you test it, still works, though

src/scry/symbol.cr
```diff
@@ -121,33 +121,35 @@ module Scry
...
-        symbols.concat search_symbols(@workspace_files, @query)
-        symbols.concat @@crystal_path_symbols.select(&.name.match(Regex.new(@query)))
+        symbols.concat search_symbols(@workspace_files, @query_regex)
+        symbols.concat @@crystal_path_symbols.select(&.name.match(@query_regex))
       end
       ResponseMessage.new(@msg_id, symbols)
     end
 
-    def search_symbols(files, query)
+    def search_symbols(files, query_regex)
```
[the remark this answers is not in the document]
faustinoaq: Done!

src/scry/symbol.cr
```diff
@@ -120,28 +120,31 @@ module Scry
...
-      end
       @query_regex = Regex.new(@query)
     end
 
     def run
       symbols = [] of SymbolInformation
       unless @query.empty?
-        symbols.concat search_symbols(@workspace_files, @query_regex)
-        symbols.concat @@crystal_path_symbols.select(&.name.match(@query_regex))
+        symbols.concat self.class.search_symbols(@workspace_files, @query_regex)
```
bew: Weird I thought it was simpler to call..

Another way is `WorkspaceSymbolProcessor.search_symbols` or `{{@type}}.search_symbols` (as you like, `self.class` is fine too)

faustinoaq: I think I prefer `self.class` :sweat_smile:

bew: no prob'!

spec/scry/symbol_spec.cr
bew: `Use File.join` ? :smiley:

src/scry/symbol.cr
```diff
@@ -113,6 +114,49 @@ module Scry
...
+      @@crystal_path_symbols ||= begin
+        crystal_path_files = Dir.glob(File.join(Scry.default_crystal_path, "**", "*.cr"))
+        search_symbols(crystal_path_files, Regex.new(".*"))
+      end
+    end
+
+    def self.search_symbols(files, query_regex)
+      symbols = [] of SymbolInformation
+      files.each do |file|
+        visitor = SymbolVisitor.new("file://#{file}")
```
bmulvihill: I wonder if there is anyway we can reuse the SymbolProcessor here since the code is so similar?  Although it would probably need to be decoupled from the ResponseMessage, maybe we can refactor that later.

PR #107, opened 2018-05-06
Add hover

### LINKED ISSUE #103: Support hover feature
We can get context types using `crystal tool context`

I think would be nice to implement hover feature just like vscode-crystal-lang is already doing, see: https://github.com/crystal-lang-tools/vscode-crystal-lang/wiki/Settings#hover

faustinoaq:
Fixes #103 :tada: 

~~Depends on #95~~ Merged! :tada:  

vokoscreen-2018-05-05_20-30-04

This uses `crystal tool context` to get context information

### DIFF
--- a/src/scry/hover_provider.cr
+++ b/src/scry/hover_provider.cr
@@ -73,8 +73,10 @@ module Scry
       contexts.each_with_index do |context, i|
         contents << "**Context #{i + 1}**\n"
         contents << "```crystal"
+        max_size = context.keys.max_by(&.size).size
         context.each do |key, value|
-          contents << "#{key} : #{value}"
+          key_aligned = "#{key}".ljust(max_size)
+          contents << "#{key_aligned} : #{value}"
         end
         contents << "```"
       end
--- a/src/scry/hover_provider.cr
+++ b/src/scry/hover_provider.cr
@@ -109,37 +109,45 @@ module Scry
     # | :--- | :--- | :--- |
     # | arg1 | `String` | `String` |
     # | arg2 | `Int32 ∣ Nil` | `Bool` |
-    private def horizontal_align(contexts)
+    def horizontal_align(contexts)
       contents = IO::Memory.new
-      contents << "| Expression |"
-      contexts.each_with_index do |context, i|
-        if contexts.size > 1
+
+      # Header
+      if contexts.size == 1
+        contents << "| Expression | Type |"
+      else
+        contents << "| Expression |"
+        contexts.size.times do |i|
           contents << " Type #{i + 1} |"
-        else
-          contents << " Type |"
         end
       end
-      contents << "\n"
+      contents.puts
+
+      # Lines
       contents << "| :--- |"
-      contexts.each_with_index do |context, i|
-        contents << " :--- |"
-      end
-      contents << "\n"
-      contents_hash = {} of String => Array(String)
-      contexts.each_with_index do |context, i|
-        context.each do |var_name, var_type|
+      contexts.size.times { contents << " :--- |" }
+      contents.puts
+
+      # Types
+      # > group var types (for all context) for each var
+      types_per_var = {} of String => Array(String)
+      contexts.each do |ctx|
+        ctx.each do |var_name, var_type|
           # Replaces | by unicode symbol 0x2223 (DIVIDES)
+          # as | can conflicts with the markdown table
           escaped_var_type = "`#{var_type.gsub("|", "∣")}`"
-          if contents_hash[var_name]?
-            contents_hash[var_name] << escaped_var_type
+          if types = types_per_var[var_name]?
+            types << escaped_var_type
           else
-            contents_hash[var_name] = [escaped_var_type]
+            types_per_var[var_name] = [escaped_var_type]
           end
         end
       end
-      contents_hash.each do |var_name, var_type|
-        contents << "| #{var_name} | #{var_type.join(" | ")} |\n"
+      # > print types, var by var
+      types_per_var.each do |var_name, var_type|
+        contents.puts "| #{var_name} | #{var_type.join(" | ")} |"
       end
+
       contents.to_s
     end
   end
--- a/src/scry/hover_provider.cr
+++ b/src/scry/hover_provider.cr
@@ -63,13 +63,12 @@ module Scry
     end
 
     private def response_with(contexts, range)
-      contents = ["```crystal"]
+      contents = ["| Expression | Type |", "| --- | --- |"]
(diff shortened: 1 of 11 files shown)


faustinoaq:
screenshot_20180505_211014

Makdown content has been enhanced :sparkles:

src/scry/hover_provider.cr
```diff
@@ -63,13 +63,47 @@ module Scry
...
+      contents = [] of String
+      titles = ["| Expression |"] of String
+      titles_lines = ["| :--- |"] of String
+      contents_hash = {} of String => Array(String)
+      contexts.each_with_index do |context, i|
+        titles << " Type #{i + 1} |"
+        titles_lines << " :--- |"
         context.each do |key, value|
-          contents << "| #{key} | #{value} |"
+          escaped_value = "`#{value.gsub("|", "∣")}`"
```
bew: is that a special unicode pipe? if yes, can you add a comment saying that (and maybe the unicode code (in hexa) too)

src/scry/hover_provider.cr
```diff
@@ -63,13 +63,47 @@ module Scry
...
     private def response_with(contexts, range)
-      contents = ["| Expression | Type |", "| --- | --- |"]
-      contexts.each do |context|
+      content = vertical_align(contexts)
+      # content = horizontal_align(contexts)
+      hover_response(Hover.new(MarkupContent.new("markdown", content.join("\n")), range))
+    end
+
+    private def vertical_align(contexts)
+      contents = [] of String
```
bew: Use an `IO::Memory` instead?

src/scry/hover_provider.cr
```diff
@@ -63,13 +63,47 @@ module Scry
...
+      contexts.each_with_index do |context, i|
+        contents << "**Context #{i + 1}**\n"
```
bew: If there is only one context we should hide the `Context 1/2/...`

src/scry/hover_provider.cr
```diff
@@ -63,13 +63,47 @@ module Scry
...
+    end
+
+    # NOTE: this Would be configurable in the future with scry.yml
+    private def horizontal_align(contexts)
+      contents = [] of String
+      titles = ["| Expression |"] of String
+      titles_lines = ["| :--- |"] of String
+      contents_hash = {} of String => Array(String)
+      contexts.each_with_index do |context, i|
+        titles << " Type #{i + 1} |"
```
bew: ditto

src/scry/hover_provider.cr
faustinoaq: Done!

src/scry/hover_provider.cr
faustinoaq: Done!

src/scry/hover_provider.cr
faustinoaq: Done!

src/scry/hover_provider.cr
faustinoaq: Done!

faustinoaq:
screenshot_20180506_000708

Changed to vertical layout :tada:

src/scry/hover_provider.cr
```diff
@@ -0,0 +1,137 @@
...
+    # ```crystal
+    # arg1 : String
+    # arg2 : Bool
+    # ```
+    private def vertical_align(contexts)
+      contents = IO::Memory.new
+      contexts.each_with_index do |context, i|
+        contents << "**Context #{i + 1}**\n" if contexts.size > 1
+        contents << "```crystal\n"
+        max_size = context.keys.max_by(&.size).size
```
bew: can be calculated outside the `each_with_index` loop

src/scry/hover_provider.cr
```diff
@@ -0,0 +1,137 @@
...
+        context.each do |key, value|
+          key_aligned = "#{key}".ljust(max_size)
```
bew: `key` is already a String, no?
So: `key.ljust(max_size)`

src/scry/hover_provider.cr
```diff
@@ -0,0 +1,137 @@
...
+    # arg1 : String
+    # arg2 : Bool
+    # ```
+    private def vertical_align(contexts)
+      contents = IO::Memory.new
+      contexts.each_with_index do |context, i|
+        contents << "**Context #{i + 1}**\n" if contexts.size > 1
+        contents << "```crystal\n"
+        max_size = context.keys.max_by(&.size).size
+        context.each do |key, value|
```
bew: `key` => `var_name`
`value` => `var_type` (or juste `type` ?)

wdyt?

src/scry/hover_provider.cr
```diff
@@ -0,0 +1,137 @@
...
+    end
+
+    # Aligns context output horizontally. By example:
+    #
+    # | Expression | Type 1 | Type 2 |
+    # | :--- | :--- | :--- |
+    # | arg1 | `String` | `String` |
+    # | arg2 | `Int32 ∣ Nil` | `Bool` |
+    private def horizontal_align(contexts)
+      contents = [] of String
```
bew: You could build this one with an `IO::Memory` too, by generating the output line by line (first the header, then the lines, then the types)

I think it would lead to cleaner/readable code

faustinoaq: Done!

src/scry/hover_provider.cr
faustinoaq: Done!

src/scry/hover_provider.cr
faustinoaq: Done!

src/scry/hover_provider.cr
faustinoaq: Done!

src/scry/protocol/hover.cr
```diff
@@ -0,0 +1,13 @@
+require "json"
+
+module Scry
+  struct Hover
+    JSON.mapping({
+      contents: MarkupContent,
+      range:    Range?,
+    }, true)
```
bew: Use named arg `static: true` here, 'cause by just reading `true` I have no idea what it refers to for a json mapping, I think it's better with the named arg

bew: Also, why only this mapping have this static check?

bmulvihill:
@faustinoaq More of an overall comment, I noticed a lot of the code between `hover_provider.cr` and `implementations.cr` is duplicated.  I am wondering if there is an shared abstraction there since they are both running the `crystal tool`

PR #114, opened 2018-05-06
Add symbol end_position using end_location

faustinoaq:
I just found this enhancement during development of rename feature

This will allow us to show code outline properly :tada: 

screenshot_20180506_163009

### DIFF
--- a/src/scry/symbol.cr
+++ b/src/scry/symbol.cr
@@ -87,20 +87,24 @@ module Scry
 
     def process_node(node, name : String, kind : SymbolKind)
       location = node.location
-      end_location = node.end_location
-      return true unless location && end_location
+      return true unless location
 
       line_number = location.line_number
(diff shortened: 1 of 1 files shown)


faustinoaq:
screenshot_20180506_163506

This is awesome :heart: almost API search with code outline :sparkles:

src/scry/symbol.cr
```diff
@@ -91,13 +91,20 @@ module Scry
 
       line_number = location.line_number
       column_number = location.column_number
-
       position = Position.new(line_number - 1, column_number - 1)
-      range = Range.new(position, position)
-      location = Location.new(@document_uri, range)
+      end_location = node.end_location
+      if end_location
```
bew: you can do `if end_location = node.end_location`, I think it shows better that `end_location` won't be used outside the if

Issue #67, opened 2018-03-22
No space left on device - reduce logger level


bew:
I usually work with _large_ files (300-400 lines), and the logging file can become very big and take all the space available in `/tmp` (which is a tmpfs on my system, so it lives in the ram).

A sample list of scry's log files:
```
-rw-------  1 lesell_b users  3.1G Mar 21 20:25 scry.out.27K0Q6
-rw-------  1 lesell_b users  1.1M Mar 21 17:24 scry.out.DJW0Dc
-rw-------  1 lesell_b users 1014K Mar 21 17:22 scry.out.KqqV1g
-rw-------  1 lesell_b users  6.8M Mar 21 23:31 scry.out.lyxMHz
-rw-------  1 lesell_b users 1014K Mar 21 17:23 scry.out.M8PRUt
-rw-------  1 lesell_b users 1013K Mar 21 17:22 scry.out.TShzqU
-rw-------  1 lesell_b users 1014K Mar 21 17:21 scry.out.VkHbyS
-rw-------  1 lesell_b users  845M Mar 21 23:27 scry.out.XTtliD
-rw-------  1 lesell_b users 1020K Mar 21 17:21 scry.out.ZtJysi
```

I suggest to _at least_ disable the logging of every request/response unless the `--debug` flag is given to scry, this will reduce the size of log files significantly I think.

faustinoaq:
Hi, @bew Scry already reduce the amount of logging by using `Log.logger.debug` when `log_level == "debug"`, Are you able to setup this in your editor?

By example in VSCode I use `crystal-lang.logLevel = "debug"` (1) when I want to get full logs otherwise normal logs should looks like:

```
-rw-------  1 main main 21K Mar 20 14:38 scry.out.fYKWU7
-rw-------  1 main main 22K Mar 20 15:43 scry.out.gJ2tFR
-rw-------  1 main main 37K Mar 20 16:24 scry.out.PoePdH
-rw-------  1 main main 27K Mar 20 17:39 scry.out.J4kKR0
-rw-------  1 main main 18K Mar 20 19:56 scry.out.4E4QX0
-rw-------  1 main main 30K Mar 20 21:45 scry.out.5ek97p
-rw-------  1 main main 15K Mar 20 22:21 scry.out.XW0lyb
-rw-------  1 main main 14K Mar 20 23:32 scry.out.9Qqf80
-rw-------  1 main main 47K Mar 21 10:34 scry.out.jPzZ7E
[...]

PR #115, opened 2018-05-08
Log to LSP client

bmulvihill:
This PR will change the logging to log to the LSP client instead of to the `scry.out` tempfile.

I am opening the PR for discussion/ideas.  I still have to write specs and such.

Currently the PR does the following
- Adds `LogMessageParams` and `MessageType` to the protocol
- Renames old `MessageType` to `ProtocolMessage`
- Creates the concept of a `Client` which can be used to send messages to the LSP client
- Allows Logger to be set


faustinoaq:
screenshot_20180507_204935

screenshot_20180507_205013

Thank you, Very useful :clap: :+1: :tada:

faustinoaq:
This is very nice, although, scry have some very verbose logs like this:

https://github.com/crystal-lang-tools/scry/blob/5e5666c8c36b8c51f2126a6f9b44599d5a8e7f6c/src/scry/completion/dependency_graph.cr#L84

So, I suggest to review them and remove the extreme verbosity:

screenshot_20180421_195116

Otherwise the lsp-client/editor would freeze because too many logs :sweat_smile: 

sometimes scry generate MB of debug data:

screenshot_20180508_095151

Sometimes, even GB, see https://github.com/crystal-lang-tools/scry/issues/67 by @bew

bmulvihill:
@faustinoaq Yes I noticed that as well.  I can remove some of the more "verbose" logging if that is desirable or we could log that stuff a tempfile like we were doing (i.e. have two loggers, a scry logger and a client logger), but that might get confusing.  I am not sure how useful some of these really big logs statements are 🤔

faustinoaq:
@bmulvihill Yeah, that may be confusing. I think we should cleanup logging and just keep client logger.

src/scry/workspace.cr
```diff
@@ -57,7 +57,6 @@ module Scry
 
     def get_file(text_document : TextDocumentIdentifier)
       filename = TextDocument.uri_to_filename(text_document.uri)
-      Log.logger.debug("@open_files: #{@open_files}")
```
faustinoaq: Yeah, this one was the most verbose :+1:

faustinoaq:
Hey @bmulvihill I already tested this and is very useful and works very nice

@crystal-lang-tools/scry Maybe we can add a log spec on a new PR, WDYT? :sweat_smile:

bmulvihill:
@faustinoaq sorry I got sidetracked on that file bug last week when working on specs.  I’ll try to wrap this up tomorrow

src/scry/log.cr
```diff
@@ -1,20 +1,37 @@
...
     private def self.initialize_logger
-      tmpfile = Tempfile.new("scry.out")
-      log_file = File.open(tmpfile.path, "w")
-      logger = Logger.new(log_file)
-      logger.level = Logger::INFO
-      logger
+      logger = ClientLogger.new(Client.instance.io)
     end
 
     @@logger : Logger = initialize_logger
```
bew: No need for a method anymore, you can directly use `ClientLogger.new ...`

src/scry/log.cr
```diff
@@ -1,20 +1,37 @@
...
 
     def self.logger
       @@logger
     end
```
bew: Note: `class_getter logger` does the same thing ;)

(or even `class_property` now that you also have a setter below)

src/scry/protocol/message_type.cr
```diff
@@ -0,0 +1,8 @@
+module Scry
```
bew: Should be `Scry::Protocol` ?

I just realized that (all?) the LSP protocol is not in a `Protocol` module, as I think it should, to properly see on usage when you're using a protocol type or not (that's important!).

This can be done later in another PR though!

src/scry/client.cr
```diff
@@ -0,0 +1,56 @@
+require "./protocol/initialize"
+require "./protocol/response_message"
+require "./protocol/notification_message"
+
+module Scry
+  class Client
+    alias ClientMessage = Initialize | ResponseMessage | NotificationMessage | Nil
+
+    def self.instance(io : IO = STDOUT)
```
bew: I think this is a bit tricky if you want to use your own IO (i.e: in specs), you _need_ to make sure that the first call to `Client.instance` has an IO, and you _can't change_ the IO later once it's set.

I think it would be better to keep the `Client.instance` as a simple getter, and change the architecture: maybe `Client` doesn't even need to be a singleton?

bew:
As I said in one of the comment, I think `Client` should not be a singleton.

Here is a way to do it:

- Store the client in `ClientLogger`, and use it in `ClientLogger#write` to send the log message.
- Configure the logger in `Scry.start`, not as a default in log.cr. You'll also have to instantiate a `Client` with its IO before the logger init and pass that client to `ClientLogger` constructor.
- And finally, remove the default `io` param to `Client` constructor, to force you to be explicit!

This way you never have to use a `Client.instance` method, you always have the current `Client` object to send things. And additionally, you can configure things easily, and repeatably (for specs).

Do you see what I mean?

src/scry/protocol/message_type.cr
faustinoaq: Yeah, I'm already working on that :+1:

bmulvihill:
I am also wondering if there is any use for the `Log` class anymore.  I just left it in to maintain the interface, but it might not have much value anymore.

src/scry/log.cr
```diff
@@ -4,21 +4,15 @@ require "./protocol/log_message_params"
...
-      @@logger = logger
-    end
+    class_property logger : Logger = Logger.new(nil)
 
     class ClientLogger < Logger
+      def initialize(@client : Client, @level = Severity::INFO, @formatter = DEFAULT_FORMATTER, @progname = "")
+        @closed = false
+        @mutex = Mutex.new
+        @io = @client.io
+      end
```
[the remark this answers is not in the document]
bmulvihill: Ahh yes, good call ;)

faustinoaq:
@bmulvihill Hi, I just tried latest changes and this feature still works fine :tada: 

Although, I notices a new `null` on my logs on every request, Why? :sweat_smile: 

screenshot_20180517_220910

screenshot_20180517_220835

BTW, Do you know what happened with the `initializer` capabilities log we used to have? :sweat_smile:

bmulvihill:
@crystal-lang-tools/scry Made some adjustments, null should no longer be logged.

PR #116, opened 2018-05-09
Do not use File.each_line iterator to ensure Files are closed

bmulvihill:
Currently when running the specs on my Mac I get a `Too many files open` error.  This is because when loading the DependencyGraph we are using the `File.each_line` iterator method, which does not close the file.  This changes the DependencyGraph to use the `File.each_line` method which yields to a block, and properly closes the File.

### DIFF
--- a/src/scry/completion/dependency_graph.cr
+++ b/src/scry/completion/dependency_graph.cr
@@ -122,16 +122,17 @@ module Scry::Completion::DependencyGraph
 
     def parse_requires(file_path)
       file_dir = File.dirname(file_path)
-      File.each_line(file_path).reduce([] of String) do |acc, line|
+      paths = [] of String
+      File.each_line(file_path) do |line|
         require_statement = /^\s*require\s*\"(?<file>.*)\"\s*$/.match(line)
         unless require_statement.nil?
           path = resolve_path(require_statement.as(Regex::MatchData)["file"].not_nil!, file_dir)
           if path
-            acc << path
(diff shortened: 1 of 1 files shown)


src/scry/completion/dependency_graph.cr
```diff
@@ -122,16 +122,17 @@ module Scry::Completion::DependencyGraph
 
     def parse_requires(file_path)
       file_dir = File.dirname(file_path)
-      File.each_line(file_path).reduce([] of String) do |acc, line|
+      paths = [] of String
+      File.each_line(file_path) do |line|
         require_statement = /^\s*require\s*\"(?<file>.*)\"\s*$/.match(line)
         unless require_statement.nil?
```
bew: Note: can be written as `if require_statement`
But that's not related to this pr

bmulvihill:
@bew `File.each_line` without a block will not close the File since it delegates to the iterator method here: https://github.com/crystal-lang/crystal/blob/master/src/file.cr#L628, which in turn calls `File.open` without a block, which simply creates a new File.  Using the block, as this PR does will delegate to https://github.com/crystal-lang/crystal/blob/master/src/file.cr#L574-L581 which ensures the file is closed

bew:
Nice find! Can you open an issue on Crystal on this, I think this is a bug, or it should be documented

faustinoaq:
@bmulvihill Nice, Thank you! :+1:

PR #122, opened 2018-06-14
Only analyze twice if main file exists

faustinoaq:

I have been working on several crystal projects latest weeks and I realize sometimes scry is very slow because is launching too much heavy crystal processes in background.

This change removes the implicit requires by removing `require "./src/*"` from analyzer, you still can use this behaivior by adding the explicit require inside your `.scry_main.cr` file.

BTW, I already have some progress in `.scry.yml` configuration, so you will be able to edit the main file very soon :wink:

### DIFF
--- a/src/scry/analyzer.cr
+++ b/src/scry/analyzer.cr
@@ -28,14 +28,10 @@ module Scry
       else
         # Crystal compiler needs a main file to compile
         # By default it uses .scry_main.cr on workspace root,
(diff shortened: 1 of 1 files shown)


bew:
Thanks @faustinoaq :smiley: 

> BTW, I already have some progress in .scry.yml configuration, so you will be able to edit the main file very soon

You're talking about the config file for scry, right?

PR #124, opened 2018-06-14
Remove query_regex on workspace symbol search

faustinoaq:

This PR removes `query_regex` because using regex can be ultra slow, also can produce unexpected behaviors :sweat_smile:

### DIFF
--- a/spec/scry/symbol_spec.cr
+++ b/spec/scry/symbol_spec.cr
@@ -114,13 +114,6 @@ module Scry
         result.kind.is_a?(SymbolKind::Method).should be_true
       end
 
-      it "return stdlib Symbol with query match for File" do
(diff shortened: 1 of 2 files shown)


spec/scry/symbol_spec.cr
```diff
@@ -107,29 +107,15 @@ module Scry
         result.empty?.should be_true
       end
 
-      it "return Symbols list with query match" do
-        processor = WorkspaceSymbolProcessor.new(0, ROOT_PATH, "salut")
+      it "return Symbols list with query match for saludo (example file)" do
```
bew: typo `saludo` => `saluto`

faustinoaq: Done! Thank you! :+1:

PR #125, opened 2018-06-14
Clean file diagnostics on close

faustinoaq:

This PR add `.full_clean` method to clean file diagnostics when a file is closed, this can allow us to remove unwanted diagnostics from editor. Current behavior keeps diagnostics even if they are already fixed because we need to open it again and configure `.scry_main`. This approach is already used when a file is deleted.

### DIFF
--- a/src/scry/context.cr
+++ b/src/scry/context.cr
@@ -143,6 +143,7 @@ module Scry
       case msg.method
       when "textDocument/didClose"
         @workspace.drop_file(params)
+        return PublishDiagnostic.new(@workspace, params.text_document.uri).full_clean
       end
       nil
     end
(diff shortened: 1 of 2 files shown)


src/scry/context.cr
```diff
@@ -143,6 +143,7 @@ module Scry
       case msg.method
       when "textDocument/didClose"
         @workspace.drop_file(params)
+        return PublishDiagnostic.new(@workspace, params.text_document.uri).full_clean
```
bew: I'd better use:
```cr
case msg.method
when "textDocument/didClose"
  @workspace.drop_file(params)
  PublishDiagnostic.new(@workspace, params.text_document.uri).full_clean
else
  nil
end
```
to simplify control flow

faustinoaq: Done! :+1:

bew:
@faustinoaq Hey, I hope you're doing well :smiley: 
Can you rebase and fix the spec error? (I'm not sure if it's due to this PR or not..)

PR #132, opened 2018-08-12
Log config updates

### LINKED ISSUE #121: Initial debug logs not being output
Currently we log the messages coming from the LSP client for debugging purposes here:

https://github.com/crystal-lang-tools/scry/blob/26a1e431c3f3b1dac87adac11af6ddf5ee6ee42c/src/scry/request.cr#L43-L47

The problem is we receive some messages (i.e. initialization and user configs) prior to the log level being set by the User settings (which is how we currently change the log level).  The default log level is `INFO` so these messages will never get logged.

bmulvihill:
Closes #121 

Logs Scry config settings at info level anytime they are changed.

### DIFF
--- a/src/scry/update_config.cr
+++ b/src/scry/update_config.cr
@@ -22,6 +22,7 @@ module Scry
     def run
       @workspace.max_number_of_problems = customizations.max_number_of_problems
       Log.logger.level = log_level(customizations.log_level || "error")
+      Log.logger.info("Scry Configuration Updated:\n #{@settings.to_json}")
       {@workspace, nil}
     end
 


src/scry/update_config.cr
```diff
@@ -22,6 +22,7 @@ module Scry
     def run
       @workspace.max_number_of_problems = customizations.max_number_of_problems
       Log.logger.level = log_level(customizations.log_level || "error")
+      Log.logger.info("Scry Configuration:\n #{@settings.to_json}")
```
bew: Maybe add a mention saying that the config has been **updated** ?

PR #133, opened 2018-08-13
Reduce some code duplication for WS symbols

bmulvihill:
Attempted to cleanup some of the code duplication when determining Workspace symbols.  In the process I also removed the coupling of the `SymbolProcessors` and the actual `ResponseMessage`.  I think doing this will allow us to reuse some of this for additional Symbol features more easily.

### DIFF
--- a/spec/scry/symbol_spec.cr
+++ b/spec/scry/symbol_spec.cr
@@ -2,56 +2,50 @@ require "../spec_helper"
 
 module Scry
   describe SymbolProcessor do
-    it "returns a ResponseMessage" do
-      text_document = TextDocument.new("inmemory://model/3", [""])
-      processor = SymbolProcessor.new(text_document)
-      processor.run.is_a?(ResponseMessage).should be_true
-    end
-
     it "contains SymbolInformation" do
       text_document = TextDocument.new("uri", [""])
       processor = SymbolProcessor.new(text_document)
       response = processor.run
-      response.result.is_a?(Array(SymbolInformation)).should be_true
+      response.is_a?(Array(SymbolInformation)).should be_true
     end
 
     it "returns Class symbols" do
       text_document = TextDocument.new("uri", ["class Test; end"])
       processor = SymbolProcessor.new(text_document)
-      response = processor.run
-      result = response.result.as(Array(SymbolInformation)).first
+      symbols = processor.run
+      result = symbols.first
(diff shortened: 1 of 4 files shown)


spec/scry/symbol_spec.cr
```diff
@@ -2,81 +2,75 @@ require "../spec_helper"
...
     end
 
     describe "Property" do
       it "returns getters as Property symbols" do
         text_document = TextDocument.new("uri", ["class Foo; getter bar : Nil; end"])
         processor = SymbolProcessor.new(text_document)
-        response = processor.run
-        result = response.result.as(Array(SymbolInformation)).try { |r| r[1] }
+        symbols = processor.run
+        result = symbols.as(Array(SymbolInformation)).try { |r| r[1] }
```
bew: Are you sure you need the cast? (ditto below)

spec/scry/symbol_spec.cr
```diff
@@ -2,81 +2,75 @@ require "../spec_helper"
...
-      processor = SymbolProcessor.new(text_document)
-      processor.run.is_a?(ResponseMessage).should be_true
-    end
-
     it "contains SymbolInformation" do
       text_document = TextDocument.new("uri", [""])
       processor = SymbolProcessor.new(text_document)
       response = processor.run
-      response.result.is_a?(Array(SymbolInformation)).should be_true
+      response.is_a?(Array(SymbolInformation)).should be_true
```
bew: Rename `response` to `symbols`

spec/scry/symbol_spec.cr
bmulvihill: Yeah this is old code, I can probably clean this up as well now

bew: Yes like below, using `symbols.first` should be enough

spec/scry/symbol_spec.cr
```diff
@@ -46,32 +46,32 @@ module Scry
       processor = SymbolProcessor.new(text_document)
       symbols = processor.run
       result = symbols.first
-      result.as(SymbolInformation).kind.is_a?(SymbolKind::Variable).should be_true
+      result.kind.is_a?(SymbolKind::Variable).should be_true
```
bew: I think this works too: `result.kind.should be_a(SymbolKind::Variable)`

PR #134, opened 2018-08-17
Protocol namespace

bmulvihill:
This PR does two things to try and cleanup/organize the code some more.

- Adds the `Protocol` namespace, this should make things more clear about what is part of the LSP Protocol
- Adds `protocol.cr` and requires that in `scry.cr` instead of requiring all of the protocol files throughout the program

### DIFF
--- a/src/scry/protocol/cancel_params.cr
+++ b/src/scry/protocol/cancel_params.cr
@@ -1,6 +1,6 @@
 require "json"
 
-module Scry
+module Scry::Protocol
   struct CancelParams
     JSON.mapping({
       id: Int32 | String,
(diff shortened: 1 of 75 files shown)


src/scry/protocol/cancel_params.cr
```diff
@@ -1,9 +1,11 @@
 require "json"
 
 module Scry
-  struct CancelParams
-    JSON.mapping({
-      id: Int32 | String,
-    }, true)
+  module Protocol
```
bew: you can do blend the 2 modules to `module Scry::Protocol`, and avoid adding another indentation inside

bmulvihill: I guess thats personal preference.  I always prefer to keep module declarations separate, maybe its from my Ruby background where it was always advised to do it this way to avoid the situation where a module wasn't defined prior to writing the Path.  I think this way is alway more clear that `Protocol` is a module.

bmulvihill:
@bew on second thought I like that approach better.  This isn't Ruby after all :)

bmulvihill:
@bew any additional input regarding the namespace?

bmulvihill:
@bew need to rebase this and pull in the 0.26.0 changes

PR #137, opened 2018-09-19
Decouple Workspace from Protocol

bmulvihill:
Attempting to clean up some the code.  It seems like we should decouple the Protocol params from the actual Workspace class.  In this case we were passing in the params, and then immediately instantiating a TextDocument, which we already had available prior to calling some of these Workspace methods.  Now we will just pass in a `TextDocument`.  

This also makes me wonder if we should either rename `TextDocument` to something like `SourceFile` or possibly rename the Workspace methods to something like `put_document` just to be consistent.

### DIFF
--- a/spec/scry/workspace_spec.cr
+++ b/spec/scry/workspace_spec.cr
@@ -1,4 +1,5 @@
 require "../spec_helper"
+require "tempfile"
 
 module Scry
   describe Workspace do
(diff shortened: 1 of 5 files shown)


spec/scry/workspace_spec.cr
```diff
@@ -1,4 +1,5 @@
 require "../spec_helper"
+require "tempfile"
```
bew: Note: Tempfile will be removed in next Crystal release

bmulvihill: Yeah I made a comment in the specs that it will need to be changed w/ v0.27.0

bmulvihill:
I'm going to wait to merge until 0.27.0 is released (which should be soon), so the Tempfile stuff doesn't have to be reworked

PR #140, opened 2018-10-12
Return request id w/ error response

### LINKED ISSUE #52: Missing `id` in ResponseError
When an error happens during a RPC call, a `ResponseMessage` is returned, without the message id it's currently in processing.
https://github.com/crystal-lang-tools/scry/blob/a2ac15c6d810c319c59aebc3c92d1a7bec866613/src/scry/protocol/response_error.cr#L20-L30
This makes the Language Client for Neovim crash for the request.

I looked at the code, the important bit is https://github.com/crystal-lang-tools/scry/blob/a2ac15c6d810c319c59aebc3c92d1a7bec866613/src/scry.cr#L20-L29
At line 25, the `ResponseMessage` object will create a `ReponseError` with the given exception.
But the current request's ID we are processing is lost.

bmulvihill:
Fixes  #52 

Returns the id of the Request with a Response error.  

According to the spec we should not be returning any message for Notifications.

### DIFF
--- a/src/scry.cr
+++ b/src/scry.cr
@@ -21,12 +21,12 @@ module Scry
 
     context = Context.new
     loop do
+      content = Request.new(STDIN).read
+      request = Message.new(content).parse
       begin
-        content = Request.new(STDIN).read
-        request = Message.new(content).parse
         results = context.dispatch(request)
       rescue ex
-        results = [Protocol::ResponseMessage.new(ex)]
+        results = process_error(request, ex)
       ensure
         response = [results].flatten.compact
         client.send_message(response)
@@ -38,6 +38,15 @@ module Scry
     ) unless Log.logger.nil?
   ensure
     Log.logger.close unless Log.logger.nil?
+  end
+
+  def self.process_error(request : Protocol::RequestMessage, error)
+    Protocol::ResponseMessage.new(request.id, error)
+  end
+
+  def self.process_error(request : Protocol::NotificationMessage, error)
+    Log.logger.error(%(#{error.message}\n#{error.backtrace.join('\n')}))
+    nil
   end
 end
 
--- a/src/scry/protocol/response_message.cr
+++ b/src/scry/protocol/response_message.cr
@@ -18,7 +18,7 @@ module Scry::Protocol
     def initialize(@result)
     end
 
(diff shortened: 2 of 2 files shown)


src/scry.cr
```diff
@@ -39,6 +39,15 @@ module Scry
   ensure
     Log.logger.close unless Log.logger.nil?
   end
+
+  def self.process_error(request : Protocol::RequestMessage, error)
```
bew: I'd suggest `make_error`, you're not really processing the error here

src/scry.cr
```diff
@@ -39,6 +39,15 @@ module Scry
...
+    Protocol::ResponseMessage.new(request.id, error)
+  end
+
+  def self.process_error(request : Protocol::NotificationMessage, error)
+    Log.logger.error(%(#{error.message}\n#{error.backtrace.join('\n')}))
```
bew: It's good to log the error here, but can you also log an error when the request is a `RequestMessage`?

src/scry.cr
bmulvihill: The reason I made it `process_error` is because the Error has already happened and now we are just handling it, and depending on the type of message we do something different.  In the case of a error for a `NotificationMessage` we are not making an error message.  Maybe `handle_error` would be better?

bew: Yes `handle_error` seems good, thank you

src/scry.cr
```diff
@@ -21,12 +21,12 @@ module Scry
 
     context = Context.new
     loop do
+      content = Request.new(STDIN).read
+      request = Message.new(content).parse
```
bew: Now this can crash: what if the content is an invalid/malformed json rpc request, for example a bad json that cannot be parsed, or which does not have an id for the request

bmulvihill: Yeah I had thought about that and I am thinking about the best way to handle this.  I think further refactoring needs to happen so we can differentiate errors that happen after we receive a valid message and before.  Really we also need to be able to test the server functionality so that needs some work as well 🤔

bew: True, yes it definitely need to be separated. Maybe wrap it in a begin..rescue block that only log something for now, until we work on it more thoroughly

PR #146, opened 2019-01-12
Make it clear that Message is a factory

matthewmcgarvey:
Came across this repo when I was looking for something in Crystal to contribute to and I thought I'd dive in.  Starting from the beginning I noticed this `Message` class that was being initialized and `#parse` immediately called in all cases.  Seeing how the JSON was being handled was definitely interesting but I believe that the `Message` class would be more clear if there was just the one, class-level method.

Looking back at the history I see that this `Message.new(...).parse` is from the [very first commit](https://github.com/crystal-lang-tools/scry/commit/dffefb00e38fcb045252fcd3819cf8a7e1ba12a9?short_path=04c6e90#diff-62f71642256d6493b5a748976dc28bdaR28) when it was still called `Procedure`.

Hope to be able to contribute more soon!

### DIFF
--- a/src/scry/message.cr
+++ b/src/scry/message.cr
@@ -4,7 +4,7 @@ module Scry
 
   alias ProtocolMessage = Protocol::RequestMessage | Protocol::NotificationMessage
 
-  struct Message
+  module Message
     def self.from(json : String?)
       raise InvalidContentError.new("Expected procedure content") unless json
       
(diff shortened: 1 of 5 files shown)


src/scry/message.cr
```diff
@@ -5,15 +5,10 @@ module Scry
   alias ProtocolMessage = Protocol::RequestMessage | Protocol::NotificationMessage
 
   struct Message
```
bew: `Message` can be a module now, since it doesn't get instantiated anymore

bmulvihill:
@matthewmcgarvey Thanks for the PR.  I think this looks good, I just need to figure out why CI isn't running..

PR #147, opened 2019-01-14
Make ResponseError code configurable and handle exceptions directly

matthewmcgarvey:
As far as I can tell so far, [this](https://github.com/crystal-lang-tools/scry/blob/master/src/scry.cr#L29) is the only place where errors are being returned currently and the code is always the unknown error code.  This PR will allow for better creation of error responses in future changes.


bmulvihill:
@matthewmcgarvey Thanks for the PR!  This was on my to do list :)  I'll try to take a look today.

src/scry/protocol/response_error.cr
```diff
@@ -15,15 +15,33 @@ module Scry::Protocol
...
       message: String,
       data:    Array(String)?,
     }, true)
 
-    def initialize(@message, @data)
-      @code = ErrorCodes::UnknownErrorCode.value
+    def initialize(@message, @data = nil, @code = ErrorCodes::UnknownErrorCode)
+    end
+
+    def initialize(ex : Exception)
```
bew: I suggest to keep only one `initialize` constructor, and have multiple overload of `.new` with an `Exception` or `ProtocolException` param

matthewmcgarvey: When would I want to override one versus the other?  I come from java land so I'm not too familiar with this new vs initialize concept

bew: basically it is preferred to have one `initialize` which will initialize all instance variables (for this `initialize`, the compiler will generate a class method `new` with the same signature), then you overload the `new` for other kind of object creation (from exception instead of message/data/code)

So you'd keep the top `initialize` method, then do:
```cr
def self.new(ex : Exception)
  # delegate to the `new`, generated for the above `initialize`
  new(ex.message || "Unknown error", ex.backtrace?)
end
```

same logic for the other one

PR #150, opened 2019-01-19
textDocument/definition shortcut if local require

matthewmcgarvey:
If the selection is a local require, the path can be built manually.

current file is `/home/src/main.cr` and require is:
```crystal
require "./api/controller"
```
then the path to the definition is just `/home/src/api/controller.cr` unless I'm missing something.

Also includes logic to make sure the selection is within the quotation marks.

[Benchmark says this is at least 200x faster for local requires:](https://github.com/matthewmcgarvey/scry/tree/local-require-benchmark)

screen shot 2019-01-18 at 10 36 00 pm

---

I have made the same PR on vscode plugin as well. https://github.com/crystal-lang-tools/vscode-crystal-lang/pull/78

### DIFF
--- a/src/scry/implementations.cr
+++ b/src/scry/implementations.cr
@@ -34,8 +34,32 @@ module Scry
 
     # NOTE: compiler is a bit heavy in some projects.
     def search(filename, position)
+      local_require = get_local_require(filename, position)
+      return local_require if local_require
       scope = get_scope(@workspace.root_uri)
       analyze(filename, position, scope)
+    end
+
+    LOCAL_REQUIRE_REGEX = /^require\s+"(\.{1,2}.*?)"\s*$/
+    POSITION_0 = Protocol::Position.new(0, 0)
+    RANGE_0 = Protocol::Range.new(POSITION_0, POSITION_0)
+
+    private def get_local_require(filename, position)
+      line = @text_document.get_line(position.line)
+      if match = line.match(LOCAL_REQUIRE_REGEX)
+        requirement = match[1]
+        start_pos = line.index(requirement)
+        return unless start_pos
+        end_pos = start_pos + requirement.size - 1
+        if position.character > start_pos || position.character < end_pos
+          filepath = "#{File.expand_path(requirement, File.dirname(filename))}.cr"
+          if File.exists?(filepath)
+            uri = "file://#{filepath}"
+            location = Protocol::Location.new(uri, RANGE_0)
+            return [Protocol::ResponseMessage.new(@text_document.id, location)]
+          end
+        end
+      end
     end
 
     private def crystal_tool(filename, position, scope)
--- a/src/scry/implementations.cr
+++ b/src/scry/implementations.cr
@@ -41,8 +41,8 @@ module Scry
     end
 
     LOCAL_REQUIRE_REGEX = /^require\s+"(\.{1,2}.*?)"\s*$/
-    POSITION_0 = Protocol::Position.new(0, 0)
-    RANGE_0 = Protocol::Range.new(POSITION_0, POSITION_0)
+    POSITION_0          = Protocol::Position.new(0, 0)
+    RANGE_0             = Protocol::Range.new(POSITION_0, POSITION_0)
 
     private def get_local_require(filename, position)
       line = @text_document.get_line(position.line)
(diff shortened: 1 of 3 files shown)


src/scry/implementations.cr
```diff
@@ -34,10 +34,34 @@ module Scry
...
+        if position.character > start_pos || position.character < end_pos
+          filepath = "#{File.expand_path(requirement, File.dirname(filename))}.cr"
+          if File.exists?(filepath)
+            uri = "file://#{filepath}"
+            location = Protocol::Location.new(uri, RANGE_0)
+            return [Protocol::ResponseMessage.new(@text_document.id, location)]
+          end
+        end
+      end
+    end
```
matthewmcgarvey: If anyone has any ideas on how to clean this up, please let me know.  This is pretty ugly 😨

src/scry/protocol/response_message.cr
```diff
@@ -1,6 +1,6 @@
 module Scry::Protocol
   # Add a response type when needed
-  alias ResponseTypes = Array(TextEdit) | Array(Location) | Array(SymbolInformation) | Array(CompletionItem) | CompletionItem | Hover
+  alias ResponseTypes = Array(TextEdit) | Array(Location) | Array(SymbolInformation) | Array(CompletionItem) | CompletionItem | Hover | Location
```
matthewmcgarvey: [reference](https://microsoft.github.io/language-server-protocol/specification#textDocument_definition)

src/scry/implementations.cr
bew: Could you explain basically what you're trying to do here ?

src/scry/implementations.cr
```diff
@@ -34,10 +34,34 @@ module Scry
...
+    RANGE_0             = Protocol::Range.new(POSITION_0, POSITION_0)
+
+    private def get_local_require(filename, position)
+      line = @text_document.get_line(position.line)
+      if match = line.match(LOCAL_REQUIRE_REGEX)
+        requirement = match[1]
+        start_pos = line.index(requirement)
+        return unless start_pos
+        end_pos = start_pos + requirement.size - 1
+        if position.character > start_pos || position.character < end_pos
```
bew: `&&` instead of `||`?

And if yes, you could replace it with `start_pos < position.character < end_pos`

src/scry/implementations.cr
```diff
@@ -34,10 +34,34 @@ module Scry
...
+          filepath = "#{File.expand_path(requirement, File.dirname(filename))}.cr"
```
bew: Iirc one could write `require "./foo.cr"`, so you should first check the expanded requirement without adding the extension

bew: Maybe sth like:
```cr
# a little helper
def return_if_exists?(path)
  File.exists?(path) && path
end

expanded_req = File.expand_path(requirement, File.dirname(filename))
filepath = return_if_exists?(expanded_req) || return_if_exists?("#{expanded_req}.cr")
if filepath
  # ... 
end
```

src/scry/implementations.cr
```diff
@@ -34,10 +34,34 @@ module Scry
...
 
+    LOCAL_REQUIRE_REGEX = /^require\s+"(\.{1,2}.*?)"\s*$/
+    POSITION_0          = Protocol::Position.new(0, 0)
+    RANGE_0             = Protocol::Range.new(POSITION_0, POSITION_0)
+
+    private def get_local_require(filename, position)
+      line = @text_document.get_line(position.line)
+      if match = line.match(LOCAL_REQUIRE_REGEX)
+        requirement = match[1]
+        start_pos = line.index(requirement)
```
bew: Not needed, the information is available in `match`!
`match.byte_begin(1)`
(the return below can then be removed) 

Same for `end_pos`, it can be written: `match.byte_end(1)`

src/scry/text_document.cr
```diff
@@ -81,6 +81,11 @@ module Scry
       @text.first
     end
 
+    def get_line(line_number)
+      lines = source.split("\n")
```
bew: That can be pretty bad for perf' and memory usage (especially for big files)

Maybe something with `source.each_line.skip(line_number - 1).next` ? This will still allocate a string for each line until it finds the line..

PR #152, opened 2019-03-01
Read from Client instead of from Request

matthewmcgarvey:
It makes sense that Client should be able to "read" from `STDIN` as well as write to it

### DIFF
--- a/spec/scry/client_spec.cr
+++ b/spec/scry/client_spec.cr
@@ -6,7 +6,7 @@ module Scry
       it "reads request from io" do
         input = %(Content-Length: 5\r\n\r\nHello\r\n)
         io = IO::Memory.new(input)
-        client = Client.new(io)
+        client = Client.new(io, STDOUT)
 
         request = client.read
 
@@ -26,7 +26,7 @@ module Scry
         message = Protocol::NotificationMessage.new("textDocument/publishDiagnostics", params)
 
         io = IO::Memory.new
-        client = Client.new(io)
+        client = Client.new(STDIN, io)
         client.send_message(message)
         io.to_s[0...19].should eq("Content-Length: 268")
       end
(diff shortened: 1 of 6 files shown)


src/scry/client.cr
```diff
@@ -1,10 +1,20 @@
...
   class Client
     alias ClientMessage = Protocol::Initialize | Protocol::ResponseMessage | Protocol::NotificationMessage
 
-    getter io
+    getter io_out
+
+    def initialize(@io_in : IO, @io_out : IO)
+    end
+
+    def self.new
```
bew: I'm not sure about this default constructor, I think it's better to explicitly initialize the `Client` instead of using defaults (which appear to be right for their current use case, but can be misleading for others)

spec/scry/client_spec.cr
```diff
@@ -14,7 +26,7 @@ module Scry
         message = Protocol::NotificationMessage.new("textDocument/publishDiagnostics", params)
 
         io = IO::Memory.new
-        client = Client.new(io)
+        client = Client.new(STDIN, io)
```
bew: Ditto here and below

matthewmcgarvey:
Those tests pass locally for me.  Seems to be a thing with the CI

src/scry/log.cr
```diff
@@ -7,7 +7,7 @@ module Scry
 
     class ClientLogger < Logger
       def initialize(@client : Client)
-        super(@client.io)
+        super(@client.io_out)
```
bew:
```suggestion
        super(@client.output)
```

src/scry/request.cr
```diff
@@ -3,15 +3,12 @@ require "./headers"
...
-    def initialize(@io : IO)
-      @headers = Headers.new
-      @content = uninitialized String | Nil
-    end
+    getter headers = Headers.new
+    getter content : String?
 
-    def read
-      return @content if @content
+    def initialize(@io : IO)
```
bew: I'm wondering, do we need to store the `io` in the request? Since there's no getter for `@io` and it's not used after the initialize I think it's safe to not store it, and pass the `io` as an argument to `read_headers` & `read_content`.

Also, if you do that you can put the whole request reading in a class method `from_io(io)` that will read the headers & content then simply initialize the request with the read values.
WDYT?

bew:
Hi @matthewmcgarvey why did you close? You don't want to continue working on this or something?

matthewmcgarvey:
@bew It was one of those things I kept meaning to come back to and never did.  I'm pulled in a lot of different directions lately and am not able to work with crystal anymore.  Maybe someone else can take  this code change over?  I figured it was pretty low value

PR #156, opened 2019-05-02
Add support for annotations

KimBurgess:
Adds support for parsing of user-defined annotations (crystal-lang/crystal#6063).

I'm not overly familiar with this code base to apologies if I've butchered the implementation. Any suggestions / criticism / (constructive) abuse welcomed in the thread here prior merge.

Updated specs are passing and the build output is tested and running within [crystal-lang-tools/atom-ide-crystal](https://github.com/crystal-lang-tools/atom-ide-crystal).

### DIFF
--- a/spec/scry/symbol_spec.cr
+++ b/spec/scry/symbol_spec.cr
@@ -57,29 +57,29 @@ module Scry
       result.kind.should be_a(LSP::Protocol::SymbolKind::Property)
     end
 
-    describe "Property" do
-      it "returns getters as Property symbols" do
+    describe "Variable" do
+      it "returns getters as Variable symbols" do
         text_document = TextDocument.new("uri", ["class Foo; getter bar : Nil; end"])
         processor = SymbolProcessor.new(text_document)
         symbols = processor.run
         result = symbols[1]
-        result.kind.should be_a(LSP::Protocol::SymbolKind::Property)
+        result.kind.should be_a(LSP::Protocol::SymbolKind::Variable)
       end
 
-      it "returns setters as Property symbols" do
+      it "returns setters as Variable symbols" do
         text_document = TextDocument.new("uri", ["class Foo; setter bar : Nil; end"])
         processor = SymbolProcessor.new(text_document)
         symbols = processor.run
         result = symbols[1]
-        result.kind.should be_a(LSP::Protocol::SymbolKind::Property)
+        result.kind.should be_a(LSP::Protocol::SymbolKind::Variable)
       end
 
-      it "returns properties as Property symbols" do
+      it "returns properties as Variable symbols" do
         text_document = TextDocument.new("uri", ["class Foo; property bar : Nil; end"])
         processor = SymbolProcessor.new(text_document)
         symbols = processor.run
         result = symbols[1]
-        result.kind.should be_a(LSP::Protocol::SymbolKind::Property)
+        result.kind.should be_a(LSP::Protocol::SymbolKind::Variable)
       end
     end
 
--- a/spec/scry/symbol_spec.cr
+++ b/spec/scry/symbol_spec.cr
@@ -49,6 +49,14 @@ module Scry
       result.kind.should be_a(LSP::Protocol::SymbolKind::Variable)
     end
 
+    it "returns annotation as Property symbols" do
+      text_document = TextDocument.new("uri", ["annotation Test; end"])
+      processor = SymbolProcessor.new(text_document)
+      symbols = processor.run
+      result = symbols.first
+      result.kind.should be_a(LSP::Protocol::SymbolKind::Property)
+    end
+
     describe "Property" do
       it "returns getters as Property symbols" do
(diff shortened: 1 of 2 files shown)


KimBurgess:
Excuse the force push - rebased onto the latest to include #158.

bew:
It looks good, thank you!

I would like to wait for #160 before merging this though.

KimBurgess:
Sounds good! I'll pull in those changes and update types to point to their shiny new `LSP` equivalents when that merges.

I've subscribe to that PR, but feel free to give me a prod here when it's ready if I look like I've missed it.

src/scry/protocol/symbol_information.cr
```diff
@@ -18,6 +18,7 @@ module Scry::Protocol
     Number      = 16
     Boolean     = 17
     Array       = 18
+    Annotation  = 19
```
bmulvihill: These values are defined by the LSP specification here: https://microsoft.github.io/language-server-protocol/specification#textDocument_documentSymbol

I do not believe there is an `Annotation` symbol type, we may have to use a different type or "bucket" it under an existing one.

bmulvihill: Looks like one of the Java LSP implementation maps their annotations to Properties: 
https://github.com/eclipse/eclipse.jdt.ls/blob/76934e4d7e9ff91d4000df8b4eaa3b942aef2311/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/handlers/DocumentSymbolHandler.java#L246

KimBurgess: There was a PR to add it (microsoft/language-server-protocol#131) but this looks to have been closed in favour of further discussion around standardisation of symbols kinds (microsoft/language-server-protocol#344).

That discussion has already been running for ~2 years so its probably worth while using an interim mapping as you suggest. I'm not sure if the `Property` kind is the best fit semantically though. Given LSP springs from Microsoft's naming conventions, that term normally maps to members that expose access to fields on an object (see https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/properties, https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/members/properties). From a quick look through other language server implementations `SymbolKind::Class` also gets some use for this purpose, but there doesn't appear to be any form of consensus.

@bmulvihill as one of the existing contributors for scry, what's your preference?

bew:
continuing the discussion started in [line comment](https://github.com/crystal-lang-tools/scry/pull/156#discussion_r321236231)

Here is the updated link to the specification's documentation: https://microsoft.github.io/language-server-protocol/specifications/specification-3-14/#textDocument_documentSymbol

I like `Property` for annotations, same as java.

I'd suggest the following change:

| ASTNode | SymbolKind |
|-|-|
| `Crystal::AnnotationDef` | `Property` |
| `Crystal::Var` | `Variable` (was property) |

bew:
By the way @KimBurgess you can now merge master in, LSP protocol definition is now in its own shard :)

KimBurgess:
Tops. I'll map to those SymbolKinds and bring things up to date here later today.

KimBurgess:
That's rebase'd and updated now. Looks like there was a little bit of strictness lacking in the associated specs (only checking for the SymbolKind type vs a value within this). I've fixed this up as part of this PR, but happy to separate if you'd like to keep things super granular.

bew:
No it's ok thank you for the update!

I'll check it out in a few hours

spec/scry/symbol_spec.cr
```diff
@@ -14,64 +14,72 @@ module Scry
...
+      it "returns getters as Variable symbols" do
         text_document = TextDocument.new("uri", ["class Foo; getter bar : Nil; end"])
         processor = SymbolProcessor.new(text_document)
         symbols = processor.run
         result = symbols[1]
-        result.kind.should be_a(LSP::Protocol::SymbolKind::Property)
+        result.kind.should eq(LSP::Protocol::SymbolKind::Variable)
       end
 
       it "returns setters as Property symbols" do
```
bew:
```suggestion
       it "returns setters as Variable symbols" do
```

spec/scry/symbol_spec.cr
```diff
@@ -14,64 +14,72 @@ module Scry
...
         text_document = TextDocument.new("uri", ["class Foo; setter bar : Nil; end"])
         processor = SymbolProcessor.new(text_document)
         symbols = processor.run
         result = symbols[1]
-        result.kind.should be_a(LSP::Protocol::SymbolKind::Property)
+        result.kind.should eq(LSP::Protocol::SymbolKind::Variable)
       end
 
       it "returns properties as Property symbols" do
```
bew:
```suggestion
       it "returns properties as Variable symbols" do
```

bew:
Thank you @KimBurgess!!!

KimBurgess:
Not a problem. Thanks for the review!

PR #183, opened 2021-03-30
lock the proper lsp version

peterhoeg:
The lock file wasn't updated when `shard.yml` was.


maxbertinetti:
You are right, my fault.

peterhoeg:
You may want to cut a .1 release with this.

